PR #2, opened 2022-10-11
Defaults to False since no backend info available.

cqc-melf:
reopened from https://github.com/CQCL/pytket-extensions/pull/473


cqc-melf:
Please check is this is all copied correctly @Roland-djee

cqc-alec (approved):

PR #5, opened 2022-10-12
update ci workflows to run weekly

cqc-melf:


cqc-melf:
Should be the same PR as for pytket-aqt

cqc-alec (approved):

Issue #15, opened 2022-10-28
The results returned from TketJob doesn't contain circuit names


yao-cqc:
This [example](https://cqcl.github.io/pytket/manual/manual_backend.html#embedding-into-qiskit) in the manual doesn't work with pytket-quantinuum for two reasons:

The first reason is described in #14 

The second reason is that the `Grover.amplify` method uses the name of the grover circuit to retrieve the result from a `qiskit.Result` object, but the `qiskit.Result` produced by `TketJob.result()` doesn't contain this information. See this line of [code](https://github.com/Qiskit/qiskit-terra/blob/b45afea03662bf4a2905251e26cbc008bdb3c0e1/qiskit/algorithms/amplitude_amplifiers/grover.py#L335) in qiskit. (i.e. the grover method uses the `qc: QuantumCircuit` as the key to the `Result.get_counts()` method, the `get_counts` method then uses `qc.name` to find the result)

To reproduce:

```
b = QuantinuumBackend(device_name="H1-1E")
backend = TketBackend(b, b.default_compilation_pass())
pass_ = Unroller(['u1', 'u2', 'u3', 'cx'])
qinstance = QuantumInstance(backend, pass_manager=pass_)
oracle = QuantumCircuit(2)
oracle.cz(0, 1)

def is_good_state(bitstr):
    return sum(map(int, bitstr)) == 2

problem = AmplificationProblem(oracle=oracle, is_good_state=is_good_state)
grover = Grover(quantum_instance=qinstance)
result = grover.amplify(problem)
print("Top measurement:", result.top_measurement)

--> 381     raise QiskitError('Data for experiment "%s" could not be found.' % key)
    382 if len(exp) == 1:
    383     exp = exp[0]

QiskitError: 'Data for experiment "Grover circuit" could not be found.'
```

We can probably add a `header.name` field to the result returned by `TketJob.result()`

PR #18, opened 2022-11-07
Bugfix/fix bugs in TketBackend

### LINKED ISSUE #14: TketBackend doesn't work with FullyConnected architecture

yao-cqc:
This PR should fix #14 and #15

### DIFF
--- a/tests/mock_pytket_backend.py
+++ b/tests/mock_pytket_backend.py
@@ -19,8 +19,13 @@ class MockShotBackend(Backend):
         arch: Optional[Union[Architecture, FullyConnected]] = None,
         gate_set: Optional[Set[OpType]] = None,
     ):
-        """Mock shot backend for testing qiskit embedding.
-        The readout bitstring will always be all 1s."""
+        """Mock shot backend for testing qiskit embedding. This should only be used
+        in conjunction with the TketBackend. The readout bitstring will always be 1s.
+        :param arch: The backend architecture
+        :type arch: Optional[Union[Architecture, FullyConnected]]
+        :param gate_set: The supported gateset, default to {OpType.CX, OpType.U3}
+        :type gate_set: Optional[Set[OpType]]
+        """
         self._id = 0
         self._arch = arch
         if gate_set:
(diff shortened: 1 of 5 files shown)


tests/mock_pytket_backend.py
```diff
@@ -0,0 +1,82 @@
...
+from pytket.utils.outcomearray import OutcomeArray
+
+
+class MockShotBackend(Backend):
+    def __init__(
+        self,
+        arch: Optional[Union[Architecture, FullyConnected]] = None,
+        gate_set: Optional[Set[OpType]] = None,
+    ):
+        """Mock shot backend for testing qiskit embedding.
```
cqc-melf, line 22: Can you add documentation about the parameters?

tests/mock_pytket_backend.py
```diff
@@ -0,0 +1,82 @@
+from typing import Optional, Union, List, Sequence, Set, cast
```
cqc-melf, line 1: Why are you defining this in the tests directory?

tests/mock_pytket_backend.py
cqc-melf: Same for the functions below?

cqc-melf:
Have you checked if this is compatible with the qiskit examples?

tests/mock_pytket_backend.py
yao-cqc: Do you mean why the``MockShotBackend`` is defined in tests?

yao-cqc:
> Have you checked if this is compatible with the qiskit examples?

I have tested that this change is compatible with [this notebook](https://github.com/CQCL/pytket/blob/4238cbb449c9570fdc16f862c649fdadb1370904/examples/qiskit_integration.ipynb), I think this is the only notebook where ``TketBackend`` is used. Do you know any other examples I should check?

tests/mock_pytket_backend.py
yao-cqc: The `MockShotBackend` is only used for testing qiskit embedding. I'm not sure where else I can define it.

cqc-melf:
> > Have you checked if this is compatible with the qiskit examples?
> 
> I have tested that this change is compatible with [this notebook](https://github.com/CQCL/pytket/blob/4238cbb449c9570fdc16f862c649fdadb1370904/examples/qiskit_integration.ipynb), I think this is the only notebook where `TketBackend` is used. Do you know any other examples I should check?

That is great, that was the notebook I was worried about!

tests/mock_pytket_backend.py
yao-cqc: At least for now, I don't think this backend can be useful elsewhere. These methods return fake data and can cause issues if the users aren't aware of the implementation (e.g. the number of shots returned from the get_result and the number of shots requested don't match). In the future, I think it might make more sense to have a configurable ``TketSimBackend`` (with correct behaviours) defined in pytket that can be used to test various things.

cqc-melf (approved):

PR #29, opened 2022-11-24
Use qiskit runtime for sampling in IBMQBackend.

cqc-alec:
Use the runtime service for both `IBMQBackend` and `IBMQEmulatorBackend`. For the latter, use the `ibmq_qasm_simulator` instead of the `AerBackend`.

Currently no error mitigation is applied. As an enhancement we could allow the user to pass options for this.

### DIFF
--- a/pytket/extensions/qiskit/backends/ibm.py
+++ b/pytket/extensions/qiskit/backends/ibm.py
@@ -15,6 +15,7 @@
 import itertools
 import logging
 from ast import literal_eval
+from collections import Counter
 import json
 from typing import (
     cast,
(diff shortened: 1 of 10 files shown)


pytket/extensions/qiskit/backends/ibm.py
```diff
@@ -423,40 +416,45 @@ def process_circuits(
...
                             ppcirc_strs[i],
                         )
                 else:
-                    job = self._backend.run(
-                        qcs, shots=n_shots, memory=self._config.memory
-                    )
-                    jobid = job.job_id()
+                    options = Options()
+                    options.resilience_level = 0
+                    options.execution.shots = n_shots
```
yao-cqc, line 427: Looks like the the [Options](https://github.com/Qiskit/qiskit-ibm-runtime/blob/main/qiskit_ibm_runtime/options/options.py) have default settings:
```
options.optimization_level = 3
options.transpilation.skip_transpilation = False
```

I'm not sure how this will affect our compilations in terms of benchmarking and error mitigation. I also don't know whether the transpilation setting will ignore the tket qubit placement.

cqc-alec: Yes that's true. I set `resilience_level` to 0 to turn off error mitigation; I should probably also set `transpilation.skip_transpilation` to `True` to preserve TKET placement, and even `optimization_level` to 0 so that we have control over gates. In a follow-up PR we can make all these configurable.

yao-cqc (approved):
Thanks!

PR #45, opened 2022-12-20
Replace use of deprecated `numpy.bool8`

cqc-alec:


cqc-melf:
This solves the mypy issue?

cqc-alec:
> This solves the mypy issue?

Yes, mypy was failing when using numpy 1.24.

Issue #28, opened 2022-11-24
Add support for general controlled gates in converters


cqc-alec:
For example:

```python
from qiskit import QuantumCircuit, QuantumRegister
from qiskit.circuit.library.standard_gates import HGate
from pytket.extensions.qiskit.qiskit_convert import tk_to_qiskit

qr = QuantumRegister(3)
qc = QuantumCircuit(qr)
c3h_gate = HGate().control(2)
qc.append(c3h_gate, qr)
c = tk_to_qiskit(qc)
```

Currently throws a `NotImplementedError`.

We could use the `QControlBox` for this.

CalMacCQ:
Done in #50

Issue #11, opened 2022-10-19
Add CnY and CnZ gates to converters


CalMacCQ:
CnY and CnZ gates are now supported in pytket. The conversion done by `tk_to_qiskit` should be able to handle these gates.

A test should also be added that covers conversion from a `Circuit` containing CnY and CnZ operations.

CalMacCQ:
Seems qiskit doesn't support CnZ and CnY gates directly. To do these gates you need to use the `MCMT` gate and pass in the appropriate parameters.

Given this I'm not sure how useful  adding this will be at the moment.

If CnY and CnZ gates are encountered in the `tk_to_qiskit` conversion they will be implemented in terms of other supported gates after #10.

CalMacCQ:
I thought this would be tricky. Turns out to be fairly straightforward (see #50).

CalMacCQ:
closed by #50

PR #50, opened 2022-12-23
Feature/convert more controlled gates

CalMacCQ:
Redo of #49 avoiding formatting issues.

I'm adding support for general multi-controlled gates in the qiskit converters.

Note that the diff here is a bit inflated by my VScode plugin reordering the imports (forgot to disable this) hopefully it doesn't upset mypy/pylint. Let me know if I should make a fresh PR.

Changes

Fixed handling of CnRy gates. These used to be supported but after https://github.com/CQCL/pytket-qiskit/pull/10 these were decomposed by the rebase prior to conversion. Adding the CnRy gate to the target gateset should take care of this.
Added support for CnY and CnZ gates. I previously thought this would be annoying to add https://github.com/CQCL/pytket-qiskit/issues/11 - turns out its easy.
Support other controlled gates using QControlBox https://github.com/CQCL/pytket-qiskit/issues/28
Also...
4. (A suggestion) - We frequently use conditions like if type(i) == ControlledGate rather a lot. Shouldn't we be using if isinstance(i, ControlledGate) ? https://switowski.com/blog/type-vs-isinstance/

### DIFF
--- a/pytket/extensions/qiskit/qiskit_convert.py
+++ b/pytket/extensions/qiskit/qiskit_convert.py
@@ -67,6 +67,7 @@ from pytket.circuit import (  # type: ignore
     CustomGateDef,
     Bit,
     Qubit,
+    QControlBox,
 )
 from pytket._tket.circuit import _TEMP_BIT_NAME  # type: ignore
 from pytket.pauli import Pauli, QubitPauliString  # type: ignore
@@ -283,13 +284,17 @@ class CircuitBuilder:
             if type(i) == ControlledGate:
                 if type(i.base_gate) == qiskit_gates.RYGate:
                     optype = OpType.CnRy
+                elif type(i.base_gate) == qiskit_gates.YGate:
+                    optype = OpType.CnY
+                elif type(i.base_gate) == qiskit_gates.ZGate:
+                    optype = OpType.CnZ
                 else:
-                    # Maybe handle multicontrolled gates in a more general way,
-                    # but for now just do CnRy
-                    raise NotImplementedError(
-                        "qiskit ControlledGate with "
-                        + "base gate {} not implemented".format(i.base_gate)
-                    )
+                    if type(i.base_gate) in set(_known_qiskit_gate.keys()):
+                        optype = OpType.QControlBox  # QControlBox case handled below
+                    else:
+                        raise NotImplementedError(
+                            f"qiskit ControlledGate with base gate {i.base_gate} not implemented"
+                        )
             elif type(i) == PauliEvolutionGate:
                 pass  # Special handling below
             else:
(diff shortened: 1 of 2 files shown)


pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -283,13 +284,18 @@ def add_qiskit_data(self, data: "QuantumCircuitData") -> None:
...
+                elif type(i.base_gate) == qiskit_gates.ZGate:
+                    optype = OpType.CnZ
                 else:
-                    # Maybe handle multicontrolled gates in a more general way,
-                    # but for now just do CnRy
-                    raise NotImplementedError(
-                        "qiskit ControlledGate with "
-                        + "base gate {} not implemented".format(i.base_gate)
-                    )
+                    if type(i.base_gate) in set(_known_qiskit_gate.keys()):
```
cqc-alec, line 292:
```suggestion
                    if type(i.base_gate) in _known_qiskit_gate:
```

pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -500,8 +518,11 @@ def append_tk_command_to_qiskit(
     qargs = [qregmap[q.reg_name][q.index[0]] for q in args]
     if optype == OpType.CnX:
         return qcirc.mcx(qargs[:-1], qargs[-1])
-
-    # special case
+    num_controls = len(qargs) - 1
```
cqc-alec, line 521: I think assuming that the "target" gate will always be single-qubit is a bug waiting to happen: better to set this in the individual cases below.

cqc-alec (changes requested):
Just a couple of small suggestions.

pytket/extensions/qiskit/qiskit_convert.py
CalMacCQ: In fact I could just do away with the num controls variable and just do 

```
    if optype == OpType.CnY:
        return qcirc.append(qiskit_gates.YGate().control(len(qargs) - 1), qargs)
```

cqc-alec: Yes, that would be better I think.

pytket/extensions/qiskit/qiskit_convert.py
CalMacCQ: Thanks, done.

pytket/extensions/qiskit/qiskit_convert.py
CalMacCQ: Okay, done.

pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -271,79 +272,102 @@ def circuit(self) -> Circuit:
...
-            elif type(i) == PauliEvolutionGate:
-                qpo = _qpo_from_peg(i, qubits)
+            elif optype == OpType.QControlBox:
+                base_tket_gate = _known_qiskit_gate[type(instr.base_gate)]
+                params = [param_to_tk(p) for p in instr.base_gate.params]
+                n_base_qubits = instr.base_gate.num_qubits
+                sub_circ = Circuit(n_base_qubits)
+                # use base gate name for the CircBox (shows in renderer)
+                sub_circ.name = instr.base_gate.name.capitalize()
+                sub_circ.add_gate(base_tket_gate, params, list(range(n_base_qubits)))
```
CalMacCQ, line 321: Do you think that adding the `base_tket_gate` to the list given by `list(range(n_base_qubits)` is correct here? I guess I have implicitly assumed that the base gate acts on the base qubits symmetrically.

cqc-alec: I think it is correct: a qiskit `Gate` (https://qiskit.org/documentation/stubs/qiskit.circuit.Gate.html#qiskit.circuit.Gate) is abstract and doesn't act on specific qubits,

cqc-alec (approved):

docs/changelog.rst
```diff
@@ -1,6 +1,10 @@
 Changelog
 ~~~~~~~~~
 
+0.34.0 (January 2023)
```
cqc-alec, line 4:
```suggestion
Unreleased
----------
```

cqc-alec (approved):

cqc-melf (approved):

PR #51, opened 2023-01-03
Support Python {3.9, 3.10, 3.11}

cqc-alec:


setup.py
```diff
@@ -44,7 +44,7 @@
     packages=find_namespace_packages(include=["pytket.*"]),
     include_package_data=True,
     install_requires=[
-        "pytket ~= 1.10",
+        "pytket == 1.11.0rc0",
```
cqc-melf, line 47: The CI should pick that up anyway, no need to update this here?

cqc-alec: Ah yes you are right.

cqc-alec: Actually it seems not in this case -- there is _no_ matching pytket 3.11 wheel for the initial install, so the build fails.

cqc-melf: Oh, okay.

cqc-melf:
Is this now ready for review?

cqc-melf (approved):

cqc-alec:
> Is this now ready for review?

Sorry, yes.

PR #55, opened 2023-01-13
Fix (and simplify) handling of endianness and unit ordering

### LINKED ISSUE #54: Order of qubits in statevector is not lexicographic
For example:

```python
from pytket.circuit import Circuit
from pytket.extensions.qiskit import AerStateBackend

backend = AerStateBackend()

c = Circuit()
a = c.add_q_register("a", 1)
b = c.add_q_register("b", 1)
c.X(a[0])

c0 = Circuit(2).X(0)

s = backend.run_circuit(c).get_state()
s0 = backend.run_circuit(c0).get_state()

print(s)
print(s0)
```

Output:

```
[0.+0.j 1.+0.j 0.+0.j 0.+0.j]
[0.+0.j 0.+0.j 1.+0.j 0.+0.j]
```

The statevectors should be the same, but the qubits `a[0]` and `b[0]` are mapped in the wrong order.

cqc-alec:
Fixes #54 .

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -6,6 +6,8 @@ Unreleased
 
 * Handle more multi-controlled gates in ``tk_to_qiskit`` and ``qiskit_to_tk`` converters (including CnY and CnZ).
 * Drop support for Python 3.8; add support for 3.11.
+* Fix ordering of registers in statevector simulation results.
+* Remove ``reverse_index`` argument in ``tk_to_qiskit()``.
 
 0.33.0 (December 2022)
 ----------------------
--- a/pytket/extensions/qiskit/backends/aer.py
+++ b/pytket/extensions/qiskit/backends/aer.py
@@ -160,14 +160,12 @@ class _AerBaseBackend(Backend):
         handle_list: List[Optional[ResultHandle]] = [None] * len(circuits)
         circuit_batches, batch_order = _batch_circuits(circuits, n_shots_list)
 
-        reverse_index = replace_implicit_swaps = (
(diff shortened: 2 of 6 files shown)


CalMacCQ:
I notice that you removed the `reverse_index` parameter. I guess all this was doing was deciding whether the `QuantumCircuit.reverse_bits()` method was used or not and can probably be left to the user.

I wonder if we should have an analogous reversal method on the `Circuit` class. I.e. given a `QuantumCircuit` it should be easy to get a pytket `Circuit` that corresponds to the same unitary matrix.

I think I should add a short manual snippet clarifying qubit ordering in qiskit vs tket.

cqc-melf (approved):

cqc-alec:
> I notice that you removed the `reverse_index` argument in `tk_to_qiskit` . I suppose all this was doing was deciding whether the `QuantumCircuit.reverse_bits()` method was used or not and can probably be left to the user.
> 
> I wonder if we should have an analogous reversal method on the `Circuit` class. I.e. given a `QuantumCircuit` it should be easy to get a pytket `Circuit` that corresponds to the same unitary matrix. (**Edit:** although I guess the user could just use `QuantumCircuit.reverse_bits()` for both directions).
> 
> Maybe I should add more to the manual clarifying qubit ordering in qiskit vs tket.

I don't think the method was really useful to the user; it was only used internally to adjust a circuit in order to obtain the correct unitary at the end (since qiskit uses little-endian instead of big-endian convention for the matrix indices), but I have changed the logic there so that this is all handled in the results conversion, which I think is cleaner.

PR #74, opened 2023-02-20
update AerBackend docs and class highlighting

CalMacCQ:
I've updated the documentation of `AerBackend` and added some highlighting for the different python classes in the index pages.

Let me know if you think more should be added. It might be nice to explain more about how `AerBackend` relates to the IBM simulator and maybe show how to define a `NoiseModel` or custom architecture.

### DIFF
--- a/docs/intro.txt
+++ b/docs/intro.txt
@@ -16,13 +16,36 @@ Windows. To install, run:
 
     pip install pytket-qiskit
 
-This will install `pytket` if it isn't already installed, and add new classes
-and methods into the `pytket.extensions` namespace.
+This will install ``pytket`` if it isn't already installed, and add new classes
+and methods into the ``pytket.extensions`` namespace.
+
+An example using the shots based :py:class:`AerBackend` simulator is shown below. 
+
+::
+
+  from pytket.extensions.qiskit import AerBackend
+
+  backend = AerBackend()
+  circ = Circuit(2).H(0).CX(0, 1).measure_all()
+
+  # Compilation not needed here as both H and CX are supported gates
+  result = backend.run_circuit(circ, n_shots=1000)
+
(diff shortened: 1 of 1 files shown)


docs/intro.txt
```diff
@@ -16,13 +16,36 @@ Windows. To install, run:
 
     pip install pytket-qiskit
 
-This will install `pytket` if it isn't already installed, and add new classes
-and methods into the `pytket.extensions` namespace.
+This will install ``pytket`` if it isn't already installed, and add new classes
+and methods into the ``pytket.extensions`` namespace.
+
+An example using the shots based :py:class:`AerBackend` simulator is shown below.
```
cqc-alec, line 22:
```suggestion
An example using the shots-based :py:class:`AerBackend` simulator is shown below. 
```

docs/intro.txt
```diff
@@ -16,13 +16,36 @@ Windows. To install, run:
...
+  from pytket.extensions.qiskit import AerBackend
+
+  backend = AerBackend()
+  backend._backend.set_option("device", "GPU")
 
 Access and Credentials
 ======================
 
-Accessing devices and simulators through the pytket-qiskit extension requires an IBMQ account. An account can be set up here -> https://quantum-computing.ibm.com/login.
+With the exception of the Aer simulators accessing devices and simulators through the ``pytket-qiskit`` extension requires an IBMQ account. An account can be set up here -> https://quantum-computing.ibm.com/login.
```
cqc-alec, line 48:
```suggestion
With the exception of the Aer simulators, accessing devices and simulators through the ``pytket-qiskit`` extension requires an IBMQ account. An account can be set up here: https://quantum-computing.ibm.com/login.
```

cqc-alec (changes requested):
Just a couple of tiny suggestions.

cqc-alec:
Use of `IBMQ` is already deprecated (in favour of `QiskitRuntimeService`) and will soon be removed -- at that point we will have to rewrite some of the authentication machinery and update these docs. But we'll cross that bridge when we come to it.

PR #76, opened 2023-02-24
Fix order of Pauli terms

cqc-alec:
Bug reported by Iakov, who has also tested the fix. Problem is difference of endianness convention between pytket and qiskit.

### DIFF
--- a/pytket/extensions/qiskit/backends/aer.py
+++ b/pytket/extensions/qiskit/backends/aer.py
@@ -696,7 +696,9 @@ def _qubitpauliop_to_sparsepauliop(
     for term, coeff in operator._dict.items():
         termmap = term.map
         strings.append(
-            "".join(termmap.get(Qubit(i), Pauli.I).name for i in range(n_qubits))
+            "".join(
+                termmap.get(Qubit(i), Pauli.I).name for i in reversed(range(n_qubits))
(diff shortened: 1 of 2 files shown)


pytket/extensions/qiskit/backends/aer.py
```diff
@@ -696,7 +696,9 @@ def _qubitpauliop_to_sparsepauliop(
     for term, coeff in operator._dict.items():
         termmap = term.map
         strings.append(
-            "".join(termmap.get(Qubit(i), Pauli.I).name for i in range(n_qubits))
+            "".join(
```
cqc-melf, line 699: Do we want to have some testcase for this?

cqc-alec: That would be good, though I wasn't sure how best to make a simple test (so I lazily asked Iakov). This method is onlt exposed through `get_operator_expectation_value()`...

cqc-melf (approved):

pytket/extensions/qiskit/backends/aer.py
cqc-alec: Iakov may be able to provide us with a test case. Meanwhile I will merge this so that we can do a hotfix.

Issue #100, opened 2023-04-11
Handle qiskit state preperation circuits in `qiskit_to_tk` converter


CalMacCQ:
Often if we try to convert qiskit circuits which use state preperation gadgets (i.e using `QuantumCircuit.initialize`) we will need to use `QuantumCircuit.decompose()` repeatedly in order to avoid a key error in the conversion. This is because the state preperation uses higher level circuit operations which are not supported by the  `qiskit_to_tk` converter.

We now have the `StatePreparationBox` implemented in TKET as well as multiplexor operations. So we should in theory be able to handle the operations used in state prep directly in the converter without decomposing first.

CalMacCQ:
Added support for this in https://github.com/CQCL/pytket-qiskit/pull/109 .

PR #109, opened 2023-05-12
handle state prep boxes in qiskit converters

CalMacCQ:
Relates to issue https://github.com/CQCL/pytket-qiskit/issues/100

Adding handling of qiskit circuits with `Initialize` and `StatePreparation` operations in the converters using the new `StatePreparationBox`

Previously users would have had to use `QuantumCircuit.decompose()` multiple times for this to work.

I've decided not to add logic for handling multiplexors as this would be more complicated and probably of limited value for the user.

[Initialize](https://qiskit.org/documentation/stubs/qiskit.circuit.QuantumCircuit.initialize.html) inserts resets whereas [StatePreparation](https://qiskit.org/documentation/stubs/qiskit.circuit.library.StatePreparation.html) does not.

There are also several ways of constructing these (str, list, int) which I've tried to handle.

I have not added support for creating a qiskit `StatePreparation` instruction with a qiskit `Statevector` object. Let me know if I should add this.
EDIT: Turns out the `Statevector` case is handled by what I've done already.

I've also used `isinstance(a, mytype)` rather than using `type(a) == mytype` as apparently this is better.
https://switowski.com/blog/type-vs-isinstance/

### DIFF
--- a/pytket/extensions/qiskit/qiskit_convert.py
+++ b/pytket/extensions/qiskit/qiskit_convert.py
@@ -57,6 +57,7 @@ from qiskit.circuit.library import CRYGate, RYGate, PauliEvolutionGate, StatePre
 
 from qiskit.extensions.unitary import UnitaryGate  # type: ignore
 from qiskit.extensions import Initialize  # type: ignore
+from qiskit.quantum_info import Statevector 
 from pytket.circuit import (  # type: ignore
     CircBox,
     Circuit,
@@ -258,7 +259,7 @@ def _string_to_circuit(
         for qubit in circ.qubits:
             circ.add_gate(OpType.Reset, [qubit])
 
-    for count, char in enumerate(circuit_string):
+    for count, char in enumerate(reversed(circuit_string)):
         if char == "0":
             pass
         if char in ("1", "-"):
@@ -376,7 +377,8 @@ class CircuitBuilder:
                         pytket_state_prep_box = StatePreparationBox(
                             amplitude_list, with_initial_reset=False
                         )
-                    self.tkc.add_gate(pytket_state_prep_box, qubits)
+                    reversed_qubits = list(reversed(qubits))
+                    self.tkc.add_gate(pytket_state_prep_box, reversed_qubits)
 
                 elif isinstance(instr.params[0], complex) and len(instr.params) == 1:
                     # convert int to a binary string and apply X for |1>
--- a/pytket/extensions/qiskit/qiskit_convert.py
+++ b/pytket/extensions/qiskit/qiskit_convert.py
@@ -78,6 +78,8 @@ from pytket.utils import QubitPauliOperator, gen_term_sequence_circuit
 
 from pytket.passes import RebaseCustom, RemoveRedundancies  # type: ignore
 
+from pytket.circuit.display import view_browser
+
 if TYPE_CHECKING:
     from qiskit.providers.backend import BackendV1 as QiskitBackend  # type: ignore
     from qiskit.providers.models.backendproperties import (  # type: ignore
@@ -244,12 +246,16 @@ def _qpo_from_peg(peg: PauliEvolutionGate, qubits: List[Qubit]) -> QubitPauliOpe
         qpodict[QubitPauliString(qubits, qpslist)] = coeff
     return QubitPauliOperator(qpodict)
 
-def _string_to_circuit(opstring: str) -> Circuit:
+
+def _string_to_circuit(string_list: Union[str, list[str]], n_qubits: int) -> Circuit:
     """Helper function to handle strings in QuantumCircuit.initialize"""
-    circ = Circuit(len(opstring))
-    # String is reversed for endian-ness
-    for count, char in enumerate(reversed(opstring)):
-        circ.add_gate(OpType.Reset, [count])
+    "".join(string_list)
+    circ = Circuit(n_qubits)
+    # Add a reset to every single qubit regardless of string length - qiskit behavior
+    for qubit in circ.qubits:
+        circ.add_gate(OpType.Reset, [qubit])
+
+    for count, char in enumerate(string_list):
         if char == "0":
             pass
         if char in ("1", "-"):
@@ -259,9 +265,10 @@ def _string_to_circuit(opstring: str) -> Circuit:
             if char == "r":
                 circ.S(count)
             elif char == "l":
-                circ.Sdg(count)       
-    
+                circ.Sdg(count)
+
     return circ
+
 
 class CircuitBuilder:
     def __init__(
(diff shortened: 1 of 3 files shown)


pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -325,6 +328,14 @@ def add_qiskit_data(self, data: "QuantumCircuitData") -> None:
                 q_ctrl_box = QControlBox(c_box, instr.num_ctrl_qubits)
                 self.tkc.add_qcontrolbox(q_ctrl_box, qubits)
 
+            elif isinstance(instr, Initialize):
+                # Check that the Initialize object is constructed with a list:
+                # https://qiskit.org/documentation/stubs/qiskit.circuit.QuantumCircuit.initialize.html
+                if isinstance(instr.params, list):
```
CalMacCQ, line 334: There is a minor complication here if `Initialize` is not constructed using a python list or numpy array.

https://qiskit.org/documentation/stubs/qiskit.circuit.QuantumCircuit.initialize.html

If Initialize is constructed using a string or an integer instead it can't easily be handled by `StatePreperationBox`. I thought it best to check for this case explictly

Maybe we should add an error message for when the `instr.params` is not a list of amplitudes

cqc-alec, line 334: As it stands this code will give incorrect results when the params are not given as a list, since no gate will be added. We should either handle the other cases, or fail with an error if we encounter them.

I think it wouldn't be too hard to handle them. For a string (such as "1++") it would just be a bunch of 1-qubit gates depending on the string. For an int it would be a bunch of X gates depending on where the 1s are in the binary representation.

tests/qiskit_convert_test.py
```diff
@@ -760,3 +760,26 @@ def test_qcontrolbox_conversion() -> None:
...
+    qc_sp = QuantumCircuit(3)
+    qc_sp.initialize(ghz_state)
+    tk_sp = qiskit_to_tk(qc_sp)
+    assert tk_sp.n_gates_of_type(OpType.StatePreparationBox) == 1
+    assert tk_sp.n_gates == 1
+    ghz_array = np.array(ghz_state)
+    assert compare_statevectors(tk_sp.get_statevector(), ghz_array)
+    # State prep with ndarray of complex amplitudes
+    qc_sp2 = QuantumCircuit(2)
+    complex_statvector = np.array([0, 1 / np.sqrt(2), -1.0j / np.sqrt(2), 0])
```
cqc-alec, line 777: Please also add a test to verify that we are handling endianness correctlty. (I am not sure which convention qiskit uses here.) Both of the test states used in the tests here are symmetric under reversal of qubits up to global phase (and `compare_statevectors()` only checks up to global phase). For example `[0, 1, 0, 0]` would be a good test.

pytket/extensions/qiskit/qiskit_convert.py
CalMacCQ: Thanks, I mistakenly thought that an error would be thrown if `instr.params` was not a list but I now realise that the code would keep running and the resulting circuit would just omit the state preparation gadget making it incorrect. 

I will see if I can handle this in a better way

tests/qiskit_convert_test.py
```diff
@@ -760,3 +760,26 @@ def test_qcontrolbox_conversion() -> None:
...
+    qc_sp2.initialize(complex_statvector, qc_sp2.qubits)
+    tk_sp2 = qiskit_to_tk(qc_sp2)
+    assert tk_sp2.n_gates_of_type(OpType.StatePreparationBox) == 1
+    assert tk_sp2.n_gates == 1
+    assert compare_statevectors(tk_sp2.get_statevector(), complex_statvector)
+    # test tket -> qiskit conversion
+    converted_qiskit_qc = tk_to_qiskit(tk_sp2)
+    assert converted_qiskit_qc.count_ops()["initialize"] == 1
```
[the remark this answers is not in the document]
CalMacCQ: Will add a check for correctness.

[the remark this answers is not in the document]
CalMacCQ: I think providing a test for correctness when the circuit contains resets would be a bit awkard currently as neither `AerStateBackend` nor `Circuit.get_statevector()` support reset operations.

yao-cqc (commented):
Assume the only difference between `StatePreparation` and `Initialize` is latter having resets, we can test the correctness of the conversions by only considering `StatePreparation`. The following is an example that produces state vectors that are directly comparable to the tket ones. 

```
def get_qiskit_statevector(qc: QuantumCircuit) -> np.ndarray:
    """Given a QuantumCircuit, use aer_simulator_statevector to compute its
     statevector, return the vector with its endianness adjusted"""
    back = Aer.get_backend("aer_simulator_statevector")
    qc.save_state()
    job = back.run(qc)
    return job.result().data()["statevector"].reverse_qargs().data
qc = QuantumCircuit(2)
qc.x(0)
tkc = Circuit(2).X(0)
assert np.allclose(get_qiskit_statevector(qc), tkc.get_statevector())
```

pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -325,6 +356,37 @@ def add_qiskit_data(self, data: "QuantumCircuitData") -> None:
...
+                        )
+                    else:
+                        pytket_state_prep_box = StatePreparationBox(
+                            amplitude_list, with_initial_reset=False
+                        )
+                    self.tkc.add_gate(pytket_state_prep_box, qubits)
+
+                elif isinstance(instr.params[0], complex) and len(instr.params) == 1:
+                    # convert int to a binary string and apply X for |1>
+                    instr.params[0] = int(instr.params[0].real)
```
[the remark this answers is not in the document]
CalMacCQ: Yeah you're right. Fixed.

yao-cqc (approved):
LGTM

pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -245,6 +246,36 @@ def _qpo_from_peg(peg: PauliEvolutionGate, qubits: List[Qubit]) -> QubitPauliOpe
...
+            pass
+        if char in ("1", "-"):
+            circ.X(count)
+        if char in ("+", "-", "r", "l"):
+            circ.H(count)
+            if char == "r":
+                circ.S(count)
+            elif char == "l":
+                circ.Sdg(count)
+
```
cqc-alec, line 275: Should raise an exception if you get an unexpected character.

pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -245,6 +246,36 @@ def _qpo_from_peg(peg: PauliEvolutionGate, qubits: List[Qubit]) -> QubitPauliOpe
...
+        if char == "0":
+            pass
+        if char in ("1", "-"):
+            circ.X(count)
+        if char in ("+", "-", "r", "l"):
+            circ.H(count)
+            if char == "r":
+                circ.S(count)
+            elif char == "l":
+                circ.Sdg(count)
```
cqc-alec, line 274: This is quite confusing; I think it would be easier to follow if we had single-option cases, with `elif`:

```python
if char == "0":
    pass
elif char == "1":
   circ.X(count)
elif ...

**Review by cqc-alec — CHANGES_REQUESTED** [review]

File: pytket/extensions/qiskit/qiskit_convert.py (hunk 5 above)
Comment (CalMacCQ) [reply pytket/extensions/qiskit/qiskit_convert.py:275]: I should've added that. Will make sure it gives an exception.

Thanks

File: pytket/extensions/qiskit/qiskit_convert.py (hunk 6 above)
Comment (CalMacCQ) [reply pytket/extensions/qiskit/qiskit_convert.py:274]: Okay will separate the cases if thats preferred.

File: pytket/extensions/qiskit/qiskit_convert.py (hunk 5 above)
Comment (CalMacCQ) [reply pytket/extensions/qiskit/qiskit_convert.py:285]: Done, thanks

File: pytket/extensions/qiskit/qiskit_convert.py (hunk 6 above)
Comment (CalMacCQ) [reply pytket/extensions/qiskit/qiskit_convert.py]: Done

File: pytket/extensions/qiskit/qiskit_convert.py [hunk 7]
```diff
@@ -264,14 +264,24 @@ def _string_to_circuit(
...
+            circ.H(count)
+        elif char == "r":
+            circ.H(count)
+            circ.S(count)
+        elif char == "l":
+            circ.H(count)
+            circ.Sdg(count)
+        else:
+            raise ValueError(
+                f"Cannot parse string for character {char}."
```
cqc-alec, line 282:
```suggestion
                f"Cannot parse string for character {char}. "
```

cqc-alec (commented):
Just a tiny comment. (Missing space in error string.)

cqc-alec (approved):

Issue #112, opened 2023-05-26
default_compilation_pass does not work new IBM device 


yoshi-qc:
New IBMQ devices ibm_seattle(433), ibm_sherbrooke(127), ibm_brisbane(127) use a new primitive gate set `{ noop X Measure RangePredicate Reset Barrier Rz Conditional SX ECR }` which is not `{OpType.CX, OpType.X, OpType.SX, OpType.Rz}`. So `default_compilation_pass` does not work.

I found that GateSetPredicate of rebase_pass does not match GateSetPredicate of required_predicates.
```python
from pytket.extensions.qiskit import IBMQBackend
ibm_backend = IBMQBackend("ibm_sherbrooke",token=ibm_token)
#ibm_backend = IBMQBackend("ibm_brisbane",token=ibm_token)
ibm_backend.required_predicates
ibm_backend.rebase_pass()
``` 
https://github.com/CQCL/pytket-qiskit/blob/bfcfbb0f18b024f42836a7933605b68ab62f1d8a/pytket/extensions/qiskit/backends/ibm.py#L449-L452

CalMacCQ:
Fixed in https://github.com/CQCL/pytket-qiskit/pull/114

PR #114, opened 2023-06-01
Support ECR gate in the gateset rebase and use pytest-rerunfailures

CalMacCQ:
Relates to the issue #112 

Newer IBM devices support the [ECR gate](https://qiskit.org/documentation/stubs/qiskit.circuit.library.ECRGate.html) instead of the CX gate.

As a workaround I've defined a `RebaseCustom` to {X, SX, Rz, ECR} rather than defining a new decomposition for `auto_rebase_pass` in TKET itself

The CX replacement I've used is the following

Screenshot 2023-06-01 at 15 56 38

* I've created a new `IBMBackend._primitive_gates` attribute which is either {X, SX, Rz, CX} , {X, SX, Rz, ECR} or the value returned by the `_tk_gateset` function depending on which gates get reported back from the API. I have also removed the `IBMQBackend._standard_gates` attribute.

* Prior to this PR we did a check if the Backend supported {X, SX, Rz, CX} in the default compilation. I've removed this as I wwant the rebase to be applied in both cases. Maybe it'd make sense to check the `GateSetPredicate` for `optimisation_level=0` instead and if its satisfied then just skip the rebase.

I've also made use of the `pytest-rerunfailures` package to rerun failing tests so hopefully we won't keep getting CI failures for transient issues. This is an alternative to removing the timeout altogether which results in very long execution times for CI jobs.

```python
@pytest.mark.flaky(reruns=3, reruns_delay=10)
def test_nshots(manila_emulator_backend: IBMQEmulatorBackend) -> None:
    for b in [AerBackend(), manila_emulator_backend]:
        circuit = Circuit(1).X(0)
        circuit.measure_all()
        n_shots = [1, 2, 3]
        results = b.get_results(b.process_circuits([circuit] * 3, n_shots=n_shots))
        assert [sum(r.get_counts().values()) for r in results] == n_shots
```

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,5 +1,8 @@
 Changelog
 ~~~~~~~~~
+0.40.0 (June 2023)(unreleased)
+------------------------------
+* Added support for the {X, SX, Rz, ECR} in the default compilation pass for :py:class:`IBMQBackend` and :py:class:`IBMQEmulatorBackend`. This is the set of gates used by some of the new IBM devices.
 
 0.39.0 (May 2023)
 -----------------
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,12 +1,9 @@
 Changelog
 ~~~~~~~~~
-0.40.0 (June 2023)(unreleased)
-------------------------------
+
+0.40.0(unreleased)
+------------------
 * Added support for the {X, SX, Rz, ECR} in the default compilation pass for :py:class:`IBMQBackend` and :py:class:`IBMQEmulatorBackend`. This is the set of gates used by some of the new IBM devices.
-
-0.40.0rc1 (unreleased)
-----------------------
-
 * IBM devices are now accessed using the `qiskit-ibm-provider <https://github.com/Qiskit/qiskit-ibm-provider>`_ instead of the deprecated :py:class:`IBMQ`. This allows the newest IBM devices and simulators to be accessed through the pytket-qiskit extension.
 * Fix to the `tk_to_qiskit` converter to prevent cancellation of redundant gates when converting to qiskit.
 * Handle qiskit circuits with :py:class:`Initialize` and :py:class:`StatePreparation` instructions in the :py:meth:`qiskit_to_tk` converter. The :py:meth:`tk_to_qiskit` converter now handles :py:class:`StatePreparationBox`.
--- a/pytket/extensions/qiskit/backends/ibm.py
+++ b/pytket/extensions/qiskit/backends/ibm.py
@@ -127,7 +127,6 @@ class NoIBMQAccountError(Exception):
 class GateSet(Enum):
     X_SX_RZ_CX = {OpType.X, OpType.SX, OpType.Rz, OpType.CX}
     X_SX_RZ_ECR = {OpType.X, OpType.SX, OpType.Rz, OpType.ECR}
-    UNKNOWN = None
 
 
 class UnsupportedGateSetException(Exception):
--- a/pytket/extensions/qiskit/backends/ibm.py
+++ b/pytket/extensions/qiskit/backends/ibm.py
@@ -123,9 +123,13 @@ class NoIBMQAccountError(Exception):
         )
 
 
+_gateset_with_ecr = {OpType.X, OpType.SX, OpType.Rz, OpType.ECR}
+_gateset_with_cx = {OpType.X, OpType.SX, OpType.Rz, OpType.CX}
+
+
 class GateSet(Enum):
-    X_SX_RZ_CX = {OpType.X, OpType.SX, OpType.Rz, OpType.CX}
-    X_SX_RZ_ECR = {OpType.X, OpType.SX, OpType.Rz, OpType.ECR}
+    X_SX_RZ_CX = _gateset_with_cx
+    X_SX_RZ_ECR = _gateset_with_ecr
 
 
 class NoRebaseException(Exception):
--- a/pytket/extensions/qiskit/backends/ibm.py
+++ b/pytket/extensions/qiskit/backends/ibm.py
@@ -28,6 +28,7 @@ from typing import (
     TYPE_CHECKING,
     Tuple,
     Union,
+    Set
 )
 from warnings import warn
 from enum import Enum
@@ -169,7 +170,7 @@ ecr_rebase = RebaseCustom(
 )
 
 
-def _get_primitive_gates(gateset: set[OpType], backend: "QiskitBackend") -> set[OpType]:
+def _get_primitive_gates(gateset: Set[OpType], backend: "QiskitBackend") -> Set[OpType]:
     if gateset >= GateSet.X_SX_RZ_CX.value:
         return GateSet.X_SX_RZ_CX.value
     elif gateset >= GateSet.X_SX_RZ_ECR.value:
--- a/pytket/extensions/qiskit/backends/ibm.py
+++ b/pytket/extensions/qiskit/backends/ibm.py
@@ -415,7 +415,13 @@ class IBMQBackend(Backend):
         # https://cqcl.github.io/pytket-qiskit/api/index.html#default-compilation
         # Edit this docs source file -> pytket-qiskit/docs/intro.txt
         if optimisation_level == 0:
-            passlist.append(self.rebase_pass())
+            if self._primitive_gates == {OpType.X, OpType.SX, OpType.Rz, OpType.CX} or {
+                OpType.X,
+                OpType.SX,
+                OpType.Rz,
+                OpType.ECR,
+            }:
+                passlist.append(self.rebase_pass())
         elif optimisation_level == 1:
             passlist.append(SynthesiseTket())
         elif optimisation_level == 2:
@@ -458,7 +464,13 @@ class IBMQBackend(Backend):
                     SynthesiseTket(),
                 ]
             )
-        passlist.extend([self.rebase_pass(), RemoveRedundancies()])
+        if self._primitive_gates == {OpType.X, OpType.SX, OpType.Rz, OpType.CX} or {
+            OpType.X,
+            OpType.SX,
+            OpType.Rz,
+            OpType.ECR,
+        }:
+            passlist.extend([self.rebase_pass(), RemoveRedundancies()])
         if optimisation_level > 0:
             passlist.append(
                 SimplifyInitial(allow_classical=False, create_all_qubits=True)
--- a/pytket/extensions/qiskit/backends/ibm.py
+++ b/pytket/extensions/qiskit/backends/ibm.py
@@ -126,16 +126,14 @@ class NoIBMQAccountError(Exception):
 class GateSet(Enum):
     X_SX_RZ_CX = {OpType.X, OpType.SX, OpType.Rz, OpType.CX}
     X_SX_RZ_ECR = {OpType.X, OpType.SX, OpType.Rz, OpType.ECR}
-    UNKNOWN = None 
+    UNKNOWN = None
 
 
-class NoRebaseException(Exception):
+class UnsupportedGateSetException(Exception):
     "Exception raised in the case that no gateset rebase is supported for an IBMBackend"
 
-    def __init__(self) -> None:
-        super().__init__(
-            "No rebase is available to convert to the supported gateset of this Backend"
-        )
+    def __init__(self, gateset) -> None:
+        super().__init__(f"Conversion to the {gateset} gateset is unsupported.")
 
 
 def _save_ibmq_auth(qiskit_config: Optional[QiskitConfig]) -> None:
@@ -169,6 +167,15 @@ ecr_rebase = RebaseCustom(
     cx_replacement=_cx_replacement_with_ecr,
     tk1_replacement=_tk1_replacement_function,
 )
+
+
+def _get_primitive_gates(gateset: set[OpType]) -> Optional[GateSet]:
+    if gateset >= GateSet.X_SX_RZ_CX.value:
+        return GateSet.X_SX_RZ_CX
+    elif gateset >= GateSet.X_SX_RZ_ECR.value:
+        return GateSet.X_SX_RZ_ECR
+    else:
+        return None
 
 
 class IBMQBackend(Backend):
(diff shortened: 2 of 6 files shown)


pytket/extensions/qiskit/backends/ibm.py
```diff
@@ -391,6 +409,8 @@ def default_compilation_pass(
         if optimisation_level == 0:
             if self._standard_gateset:
                 passlist.append(self.rebase_pass())
+            else:
```
cqc-alec, line 412: Probably we should test that the gateset is compatible here, rather than assume that it is if `self._standard_gateset` is `False` (and throw an error if not). Otherwise if a device comes along with a new gateset we'd blithely submit circuits with invalid gates.

I suggest a check similar to that for `_standard_gateset`. In fact, `_standard_gateset` is probably a bad name to use now. We could have an `Enum` of known gatesets instead (and store its value as a `Backend` attribute) instead of the current boolean.

CalMacCQ: Good point, this should be done more carefully

I want to ensure that `IBMBackend.rebase_pass()` gives the right rebase too.

pytket/extensions/qiskit/backends/ibm.py
```diff
@@ -389,8 +418,7 @@ def default_compilation_pass(
         # https://cqcl.github.io/pytket-qiskit/api/index.html#default-compilation
         # Edit this docs source file -> pytket-qiskit/docs/intro.txt
         if optimisation_level == 0:
-            if self._standard_gateset:
```
CalMacCQ, line 392: I removed this "if" statement as we need to apply the rebase regardless of whether the device supports CX or ECR.

Perhaps we should check that the circuit satisfies the `GateSetPredicate` here instead? In that case we wouldn't even need a rebase.

pytket/extensions/qiskit/backends/ibm.py
```diff
@@ -199,6 +223,11 @@ def __init__(
 
         self._standard_gateset = gate_set >= {OpType.X, OpType.SX, OpType.Rz, OpType.CX}
 
+        self._primitive_gates = (
+            GateSet.X_SX_RZ_CX.value
+            if gate_set >= {OpType.X, OpType.SX, OpType.Rz, OpType.CX}
+            else GateSet.X_SX_RZ_ECR.value
```
CalMacCQ, line 229: I'm not really happy with the syntax of this. 

If there's any suggestion for imporvement let me know

pytket/extensions/qiskit/backends/ibm.py
```diff
@@ -249,6 +290,16 @@ def _get_provider(
     def backend_info(self) -> BackendInfo:
         return self._backend_info
 
+    @property
```
CalMacCQ, line 293: Here I've used the syntax recommended in https://stackoverflow.com/questions/30585427/raise-an-exception-if-class-attribute-is-set-to-a-certain-value-after-instantiat?rq=3

Let me know if this is acceptable

pytket/extensions/qiskit/backends/ibm.py
```diff
@@ -120,6 +124,19 @@ def __init__(self) -> None:
         )
 
 
+class GateSet(Enum):
+    X_SX_RZ_CX = {OpType.X, OpType.SX, OpType.Rz, OpType.CX}
+    X_SX_RZ_ECR = {OpType.X, OpType.SX, OpType.Rz, OpType.ECR}
+    UNKNOWN = None
```
CalMacCQ, line 130: Maybe the None is redundant here now? In the `UnsupportedGateSetException` I want to give the user the unsupported OpTypes as I think this is more helpful.

pytket/extensions/qiskit/backends/ibm.py
```diff
@@ -447,9 +496,10 @@ def _result_id_type(self) -> _ResultIdTuple:
         return (str, int, int, str)
 
     def rebase_pass(self) -> BasePass:
-        return auto_rebase_pass(
-            {OpType.CX, OpType.X, OpType.SX, OpType.Rz},
-        )
+        if self._primitive_gates == GateSet.X_SX_RZ_CX.value:
+            return auto_rebase_pass(GateSet.X_SX_RZ_CX.value)
+        elif self._primitive_gates == GateSet.X_SX_RZ_ECR.value:
+            return ecr_rebase
```
cqc-alec, line 502: Why not return `ecr_rebase` if `self._primitive_gates` allows it and otherwise return `auto_rebase_pass(self._primitive_gates)`? That will then raise an exception if we get a gateset that we can't rebase to.

pytket/extensions/qiskit/backends/ibm.py
```diff
@@ -120,6 +124,18 @@ def __init__(self) -> None:
         )
 
 
+class GateSet(Enum):
+    X_SX_RZ_CX = {OpType.X, OpType.SX, OpType.Rz, OpType.CX}
+    X_SX_RZ_ECR = {OpType.X, OpType.SX, OpType.Rz, OpType.ECR}
```
cqc-alec, line 129: I don't think we need this enum at all. We are storing the primitive gates as a backend attribute -- so just use that attribute when computing `rebase_pass()`?

pytket/extensions/qiskit/backends/ibm.py
```diff
@@ -137,6 +153,31 @@ def _save_ibmq_auth(qiskit_config: Optional[QiskitConfig]) -> None:
             QiskitRuntimeService.save_account(channel="ibm_quantum", token=token)
 
 
+# Variables for defining a rebase to the {X, SX, Rz, ECR} gateset
+# See https://github.com/CQCL/pytket-qiskit/issues/112
+_cx_replacement_with_ecr = (
+    Circuit(2).X(0).SX(1).Rz(-0.5, 0).add_gate(OpType.ECR, [0, 1])
+)
+_tk1_replacement_function = get_TK1_decomposition_function(
```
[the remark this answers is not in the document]
CalMacCQ: Done

pytket/extensions/qiskit/backends/ibm.py
CalMacCQ: Done, this is an improvement

pytket/extensions/qiskit/backends/ibm.py
CalMacCQ: Done. 

I've gotten rid of the enum and used the backend attribute when possible.

For the `_get_primitive_gates` function I've used

```python
def _get_primitive_gates(gateset: Set[OpType]) -> Set[OpType]:
    if gateset >= {OpType.X, OpType.SX, OpType.Rz, OpType.CX}:
        return {OpType.X, OpType.SX, OpType.Rz, OpType.CX}
    elif gateset >= {OpType.X, OpType.SX, OpType.Rz, OpType.ECR}:
        return {OpType.X, OpType.SX, OpType.Rz, OpType.ECR}
    else:
        return gateset
```

CalMacCQ:
I think the mypy CI failure is possibly due to the fact that 1.16rc is being installed?

pytket/extensions/qiskit/backends/ibm.py
CalMacCQ: Do you think raising an excpetion in the case of an unsupported gateset is sufficent? This exception would be given when the user does the default compilation pass.

I got the impression that you also wanted an exception rasied when the user tried to initialise an `IBMQBackend` object but maybe I misunderstood.

cqc-alec: > Do you think raising an excpetion in the case of an unsupported gateset is sufficent here? This exception would be given when the user does the default compilation pass.
> 
> I got the impression that you also wanted an exception rasied when the user tried to construct an `IBMQBackend` object with an unsupported gateset but maybe I misunderstood.

We can constuct the rebase pass on initialization (and then store it as an attribute).

CalMacCQ: Store the rebase pass as an attribute? I'm not sure I follow. 

Currently the primitive gates are an attribute which is used to compute the rebase in the `rebase_pass` method.

 Should I store the output of `rebase_pass` as an attribute?

cqc-alec: That's what I'm suggesting if you want to fail on backend construction but actually I've changed my mind, I don't think this is a good idea: better to fail on compilation. That allows users to submit circuits that satisfy the gateset even if the default compilation passes don't support it.

CalMacCQ: Yeah I agree I think its better to fail on compilation as well in this case. 

Thanks

pytket/extensions/qiskit/backends/ibm.py
```diff
@@ -370,7 +396,12 @@ def default_compilation_pass(
         # https://cqcl.github.io/pytket-qiskit/api/index.html#default-compilation
         # Edit this docs source file -> pytket-qiskit/docs/intro.txt
         if optimisation_level == 0:
-            if self._standard_gateset:
+            if self._primitive_gates == {OpType.X, OpType.SX, OpType.Rz, OpType.CX} or {
```
CalMacCQ, line 399: This rather messy if statement is to avoid a test failure [here](https://github.com/CQCL/pytket-qiskit/blob/1110d9e096b4bf0ae4788d844763ee89fb1bb9d1/tests/backend_test.py#L1109)

I was wondering why the `if self._standard_gateset` check was there before.

Basically its there to ensure that the rebase is skipped in the default compilation pass if its not {X, SX, Rz, CX}.

The `test_cloud_stabiliser` test compiles a circuit that already satisfies the `GateSetPredicate` for the sim so the rebase should be skipped altogether as we can't find a valid TK1 decomposition.

This confused me for quite a while

pytket/extensions/qiskit/backends/ibm.py
```diff
@@ -370,7 +398,10 @@ def default_compilation_pass(
         # https://cqcl.github.io/pytket-qiskit/api/index.html#default-compilation
         # Edit this docs source file -> pytket-qiskit/docs/intro.txt
         if optimisation_level == 0:
-            if self._standard_gateset:
```
CalMacCQ, line 373: This check is to ensure that the rebase is skipped when compiling to the remote stabilizer simulator. Basically the simulator does not support the Rz gate meaning no TK1 replacement can be found for the gateset.

Skippping the rebase (as was done before this PR as well) prevents a test failure [here](https://github.com/CQCL/pytket-qiskit/blob/develop/tests/backend_test.py#LL1109C1-L1109C1)

pytket/extensions/qiskit/backends/ibm.py
```diff
@@ -370,7 +398,10 @@ def default_compilation_pass(
...
+            if self._supports_rz:
+                # If the Rz gate is unsupported then the rebase should be skipped
+                # This prevents an error when compiling to the stabilizer backend
```
CalMacCQ, line 403: This check is to ensure that the rebase is skipped when compiling to the remote stabilizer simulator. Basically the simulator does not support the Rz gate meaning no TK1 replacement can be found for the gateset.

Skippping the rebase (as was done before this PR as well) prevents a test failure [here](https://github.com/CQCL/pytket-qiskit/blob/develop/tests/backend_test.py#LL1109C1-L1109C1)

docs/changelog.rst
```diff
@@ -1,5 +1,8 @@
 Changelog
 ~~~~~~~~~
+0.40.0 (June 2023)(unreleased)
+------------------------------
+* Added support for the {X, SX, Rz, ECR} in the default compilation pass for :py:class:`IBMQBackend` and :py:class:`IBMQEmulatorBackend`. This is the set of gates used by some of the new IBM devices.
```
cqc-melf, line 5: Can you please combine this to one 0.40. block with the notes from the RC

tests/backend_test.py
```diff
@@ -1101,7 +1104,7 @@ def test_postprocess_emu(manila_emulator_backend: IBMQEmulatorBackend) -> None:
     assert sum(counts.values()) == 10
 
 
-@pytest.mark.timeout(None)
+@pytest.mark.flaky(reruns=3, reruns_delay=2)
```
cqc-melf, line 1107: Do you think 2 is a good delay? I thought it should be longer, but I am not sure. Have you tried if this helps?

CalMacCQ: The tests seem to pass with a delay of 2 seconds so far. If we still get failures we can up this to 5 or 10

docs/changelog.rst
CalMacCQ: Done. I've combined these under the heading `0.40.0(unreleased)` as the rebase will belong to a different release candidate than the previous changes. Therefore I think this heading should be used instead of RC1.

tests/backend_test.py
CalMacCQ: Got a failure
https://github.com/CQCL/pytket-qiskit/actions/runs/5321790504/jobs/9637347806
I'll try 10 seconds instead of 2

cqc-alec (commented):
Can we add a test that uses a backend with the new gate set to check that compilation works?

CalMacCQ:
As far as I can tell the only devices with the new gateset are ibm_seattle, ibm_brisbane and ibm_sherbrooke. 
These have 433, 127 and 127 qubits respectively. 

Would we want jobs running regularly on these large devices every time somebody pushes to pytket-qiskit? Maybe we could set up a small emulator job or something without too many qubits or shots.

Screenshot 2023-06-20 at 15 27 28

cqc-alec:
> As far as I can tell the only devices with the new gateset are ibm_seattle, ibm_brisbane and ibm_sherbrooke. These have 433, 127 and 127 qubits respectively.
> 
> Would we want jobs running regularly on these large devices every time somebody pushes to pytket-qiskit? Maybe we could set up a small emulator job or something without too many qubits or shots so it doesn't get stuck in the queue for ages.

We wouldn't need to run any jobs, just compile some circuits.

CalMacCQ:
Ah sorry I misread. Yeah I can add a test for that.

CalMacCQ:
For a compilation test I've just used `IBMQBackend` targetting `ibm_sherbrooke`. It may also be possible to test compilation using a [Fake backend](https://qiskit.org/documentation/stubs/qiskit.providers.fake_provider.FakeSherbrooke.html).

cqc-alec (approved):

PR #115, opened 2023-06-06
Fix/tk to qiskit rendundancies

CalMacCQ:
After a change made in https://github.com/CQCL/pytket-qiskit/pull/10 a `RemoveRedundancies` pass was applied in the `tk_to_qiskit` function.

This PR removes this pass as it cancels redundancies like the following

```python
from pytket import Circuit
from pytket.extensions.qiskit import tk_to_qiskit

print(tk_to_qiskit(Circuit(1).H(0).H(0))))
```

This prints an empty `QuantumCircuit` as the two Hadamards are cancelled by `RemoveRedundancies`. This is confusing for users. The `tk_to_qiskit` converter shouldn't be doing optimisation when converting to qiskit.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,7 +1,7 @@
 Changelog
 ~~~~~~~~~
-0.40.0 (June 2023)(unreleased)
-------------------------------
+0.40.0 (unreleased)
+-------------------
 * Fix to the `tk_to_qiskit` converter to prevent cancellation of redundant gates when converting to qiskit.
 
 0.39.0 (May 2023)


docs/changelog.rst
```diff
@@ -1,5 +1,8 @@
 Changelog
 ~~~~~~~~~
+0.40.0 (June 2023)(unreleased)
```
cqc-melf, line 3:
```suggestion
0.40.0 (unreleased)
-------------------

File: docs/changelog.rst [hunk 2]
```diff
@@ -1,5 +1,8 @@
...
+------------------------------
```
cqc-melf, line 4:
```suggestion

cqc-melf (changes requested):

docs/changelog.rst
CalMacCQ: Thanks, done

docs/changelog.rst
CalMacCQ: Done

cqc-melf (approved):

Issue #111, opened 2023-05-26
IBMQBackend cannot obtain ibm_seattle device information


yoshi-qc:
Using pytket-qiskit, we cannot access to ibm_seattle(433 qubits). 
```python
ibm_token = your ibm token
from qiskit import IBMQ
IBMQ.save_account(ibm_token,hub=your hub,group=your group,project=your project, overwrite=True)
from pytket.extensions.qiskit.backends.config import set_ibmq_config
set_ibmq_config(ibmq_api_token=ibm_token,hub=your hub,group=your group,project=your project)
from pytket.extensions.qiskit import IBMQBackend
list_backend=IBMQBackend.available_devices(hub=your hub,group=your group,project=your project)
for i in range(len(list_backend)):
    print(list_backend[i].device_name)
``` 
pytket-qiskit is using api "IBMQ" to obtain backend information, but there seems to be a problem with api "IBMQ" and information on ibm_seattle(433) is not being obtained. If we use `from qiskit_ibm_provider import IBMProvider` as below, we can obtain ibm_seattle(433) information.
```python
from qiskit_ibm_provider import IBMProvider
IBMProvider.save_account(token=ibm_token,overwrite=True)
provider = IBMProvider()
hub = "partner-cqc"
group = "internal"
project = "default"
backend_name = "ibm_seattle"
backend = provider.get_backend(backend_name, instance=f"{hub}/{group}/{project}")
provider.backends()
``` 

https://github.com/CQCL/pytket-qiskit/blob/bfcfbb0f18b024f42836a7933605b68ab62f1d8a/pytket/extensions/qiskit/backends/ibm.py#LL35C1-L35C1

CalMacCQ:
Should be fixed now that we have upgraded to the `IBMProvider` in https://github.com/CQCL/pytket-qiskit/pull/116 .

Issue #64, opened 2023-01-20
Remove soon-to-be-deprecated use of `qiskit.IBMQ`


cqc-alec:
See https://github.com/Qiskit/qiskit-terra/pull/8080 .

CalMacCQ:
Done in https://github.com/CQCL/pytket-qiskit/pull/116 .

PR #116, opened 2023-06-07
Feature/ibm provider

CalMacCQ:
PR to remove the use of the deprecated `IBMQ` https://github.com/CQCL/pytket-qiskit/issues/64.

Currently `IBMQ` is blocking access to the new IBM devices https://github.com/CQCL/pytket-qiskit/issues/111

I'm replacing this with the [IBMprovider](https://github.com/Qiskit/qiskit-ibm-provider) following the [migration guide](https://qiskit.org/ecosystem/ibm-provider/tutorials/Migration_Guide_from_qiskit-ibmq-provider.html).

There were some error messages like the following when using `IBMQEmulator` of the following form

```
ERROR backend_test.py::test_postprocess_emu - qiskit.providers.exceptions.BackendPropertyError: "Couldn't find the property 'T1' for qubit 0."
```

There were fixed by constructing the `NoiseModel` directly from the backend instead of going via the `AerSimulator` in ibmq_emulator.py.

```python
self._noise_model = NoiseModel.from_backend(self._ibmq._backend)
```

### DIFF
--- a/docs/intro.txt
+++ b/docs/intro.txt
@@ -78,7 +78,7 @@ locally without saving the token in pytket config:
     from qiskit_ibm_provider import IBMProvider
     from qiskit_ibm_runtime import QiskitRuntimeService
 
-    IBMProvider.save_account(token=ibm_token, overwrite=True)
+    IBMProvider.save_account(token=ibm_token)
     QiskitRuntimeService.save_account(channel="ibm_quantum", token=ibm_token)
 
 To see which devices you can access you can use the ``available_devices`` method on the :py:class:`IBMQBackend` or :py:class:`IBMQEmulatorBackend`. Note that it is possible to pass optional ``instance`` and ``provider`` arguments to this method. This allows you to see which devices are accessible through your IBM hub.
--- a/docs/intro.txt
+++ b/docs/intro.txt
@@ -50,7 +50,7 @@ The :py:class:`AerBackend` also supports GPU simulation which can be configured 
 Access and Credentials
 ======================
 
-With the exception of the Aer simulators, accessing devices and simulators through the ``pytket-qiskit`` extension requires an IBMQ account. An account can be set up here: https://quantum-computing.ibm.com/login.
+With the exception of the Aer simulators, accessing devices and simulators through the ``pytket-qiskit`` extension requires an IBM account. An account can be set up here: https://quantum-computing.ibm.com/login.
 
 Once you have created an account you can obtain an API token which you can use to configure your credentials locally.
 
@@ -60,7 +60,7 @@ Once you have created an account you can obtain an API token which you can use t
     
     set_ibmq_config(ibmq_api_token=ibm_token)
 
-This will save your IBMQ credentials locally. After saving your credentials you can access ``pytket-qiskit`` backend repeatedly without having to re-initialise your credentials.
+After saving your credentials you can access ``pytket-qiskit`` backend repeatedly without having to re-initialise your credentials.
 
 If you are a member of an IBM hub then you can add this information to ``set_ibmq_config`` as well.
 
@@ -68,27 +68,29 @@ If you are a member of an IBM hub then you can add this information to ``set_ibm
 
     from pytket.extensions.qiskit import set_ibmq_config
 
-    set_ibmq_config(ibmq_api_token=ibm_token, hub='your hub', group='your group', project='your project')
+    set_ibmq_config(ibmq_api_token=ibm_token, instance=f"{hub}/{group}/{project}")
 
 Alternatively you can use the following qiskit commands to save your credentials
 locally without saving the token in pytket config:
 
 ::
 
-    from qiskit import IBMQ
+    from qiskit_ibm_provider import IBMProvider
     from qiskit_ibm_runtime import QiskitRuntimeService
 
-    IBMQ.save_account(token=ibm_token)
+    IBMProvider.save_account(token=ibm_token, overwrite=True)
     QiskitRuntimeService.save_account(channel="ibm_quantum", token=ibm_token)
 
-To see which devices you can access you can use the ``available_devices`` method on the :py:class:`IBMQBackend` or :py:class:`IBMQEmulatorBackend`. Note that it is possible to pass ``hub``, ``group`` and ``project`` parameters to this method. This allows you to see which devices are accessible through your IBM hub.
+To see which devices you can access you can use the ``available_devices`` method on the :py:class:`IBMQBackend` or :py:class:`IBMQEmulatorBackend`. Note that it is possible to pass optional ``instance`` and ``provider`` arguments to this method. This allows you to see which devices are accessible through your IBM hub.
 
 ::
 
     from pytket.extensions.qiskit import IBMQBackend
+    from qiskit_ibm_provider import IBMProvider
 
-    backend = IBMQBackend # Initialise backend for an IBM device
-    backend.available_devices(hub='your hub', group='your group', project='your project') 
+    ibm_provider = IBMProvider()
+    backend = IBMQBackend("ibmq_belem") # Initialise backend for an IBM device
+    backend.available_devices(instance=instance=f"{hub}/{group}/{project}", provider=ibm_provider) 
 
 
 Backends Available Through pytket-qiskit
--- a/pytket/extensions/qiskit/backends/config.py
+++ b/pytket/extensions/qiskit/backends/config.py
@@ -36,7 +36,7 @@ class QiskitConfig(PytketExtConfig):
             ext_dict.get("hub", None),
             ext_dict.get("group", None),
             ext_dict.get("project", None),
-            ext_dict.get("ibmq_api_token", None),
+            ext_dict.get("ibmq_api_token", None),  # is here an issue?
         )
 
 
--- a/pytket/extensions/qiskit/backends/config.py
+++ b/pytket/extensions/qiskit/backends/config.py
@@ -36,7 +36,7 @@ class QiskitConfig(PytketExtConfig):
             ext_dict.get("hub", None),
             ext_dict.get("group", None),
             ext_dict.get("project", None),
-            ext_dict.get("ibmq_api_token", None),  # is here an issue?
+            ext_dict.get("ibmq_api_token", None),
         )
 
 
--- a/pytket/extensions/qiskit/backends/config.py
+++ b/pytket/extensions/qiskit/backends/config.py
@@ -23,9 +23,7 @@ class QiskitConfig(PytketExtConfig):
 
     ext_dict_key: ClassVar[str] = "qiskit"
 
-    hub: Optional[str]
-    group: Optional[str]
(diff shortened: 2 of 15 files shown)


tests/test-requirements.txt
```diff
@@ -2,3 +2,4 @@ pytest
 pytest-timeout ~= 1.4.2
 hypothesis
 requests_mock
+qiskit-ibm-provider ~= 0.6.0
```
yao-cqc, line 5: Why override the qiskit versions already required by pytket-qiskit?

yao-cqc (commented):
this PR https://github.com/Qiskit/qiskit-terra/pull/9845 changed some error messages. In the `test_unsupported_gateset` case

The following change is needed
```
    # (tested with qiskit 0.39.1)
    with pytest.raises(TranspilerError) as e:
        result = grover.amplify(problem)
    err_msg = "Unable to map"
```
to
```
    # (tested with qiskit 0.xx)
    with pytest.raises(TranspilerError) as e:
        result = grover.amplify(problem)
    err_msg = "Unable to translate"
```

tests/test-requirements.txt
CalMacCQ: Installing pytket-qiskit doesn't install `qiskit-ibm-provider`. This is the newer client that we are using instead of the deprecated one.

It needs to be added to `setup.py` though. I can do that in this PR

setup.py
```diff
@@ -48,6 +48,7 @@
         "qiskit ~= 0.42.1",
         "qiskit-ibm-runtime ~= 0.9.2",
         "qiskit-aer ~= 0.12.0",
+        "qiskit-ibm-provider ~= 0.5.0",
```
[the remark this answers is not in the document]
cqc-melf: Is there a reason why we should use 0.5.0 ?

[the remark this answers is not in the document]
CalMacCQ: There are some depedency issues otherwise. This is resolved by upgrading to 0.43.1 but I thought maybe you would want to do this in a separate PR.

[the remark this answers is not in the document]
cqc-melf: Okay! Yes that is something we should do separately

tests/backend_test.py
```diff
@@ -151,10 +155,8 @@ def test_measures() -> None:
 
 @pytest.mark.skipif(skip_remote_tests, reason=REASON)
 def test_noise() -> None:
```
cqc-melf, line 157: We should define the provider in the same way the backends are created for the remote tests.

pytket/extensions/qiskit/backends/config.py
```diff
@@ -36,7 +36,7 @@ def from_extension_dict(
             ext_dict.get("hub", None),
             ext_dict.get("group", None),
             ext_dict.get("project", None),
-            ext_dict.get("ibmq_api_token", None),
+            ext_dict.get("ibmq_api_token", None),  # is here an issue?
```
cqc-melf, line 39:
```suggestion
            ext_dict.get("ibmq_api_token", None),

File: pytket/extensions/qiskit/backends/ibm.py [hunk 5]
```diff
@@ -124,14 +122,16 @@ def _save_ibmq_auth(qiskit_config: Optional[QiskitConfig]) -> None:
     token = None
     if qiskit_config is not None:
         token = qiskit_config.ibmq_api_token
-    if not IBMQ.active_account():
-        if IBMQ.stored_account():
-            IBMQ.load_account()
+    if token is None and os.getenv("PYTKET_REMOTE_QISKIT_TOKEN") is not None:
```
cqc-melf, line 125: I am not sure if it is the best way to keep this in here, but I have not found a way without this. This is run locally for all users, maybe we should add a check for an additional env variable to make sure that this is only run on the CI?

CalMacCQ:
Everything seems to be passing for now.

I had to rerun it a few times to get rid of the 401 errors. I'm not sure how much of this is casued by the pytket-qiskit side. It could be server issues with the qiskit provider also.

I've kept the lines 
```
if token is None and os.getenv("PYTKET_REMOTE_QISKIT_TOKEN") is not None:
         token = os.getenv("PYTKET_REMOTE_QISKIT_TOKEN")
```

For now. This was Melf's solution to get the checks to pass. I don't have a better idea at the moment.

cqc-melf:
> Everything seems to be passing for now.
> 
> I had to rerun it a few times to get rid of the 401 errors. I'm not sure how much of this is casued by the pytket-qiskit side. It could be server issues with the qiskit provider also.
> 
> I've kept the lines
> 
> ```python
> if token is None and os.getenv("PYTKET_REMOTE_QISKIT_TOKEN") is not None:
>          token = os.getenv("PYTKET_REMOTE_QISKIT_TOKEN")
> ```
> 
> For now. This was Melf's solution to get the checks to pass. I don't have a better idea at the moment.

Can you give some more details on how much "a few" is?
I am still not very happy with all the reruns that needs to be done, I will have another look to check if there is something that we missed so far...

tests/qiskit_convert_test.py
```diff
@@ -267,17 +269,14 @@ def test_tketpass() -> None:
...
         Aer.get_backend("aer_simulator_statevector"),
         Aer.get_backend("aer_simulator"),
         Aer.get_backend("aer_simulator_unitary"),
     ]
-    if not skip_remote_tests:
-        if not IBMQ.active_account():
-            IBMQ.load_account()
-        provider = IBMQ.providers(hub="ibm-q")[0]
-        backends.append(provider.get_backend("ibmq_manila"))
+    backends.append(manila_backend._backend)  # type: ignore
```
cqc-melf, line 279: Does this still need the type ignore with the recent changes?

tests/qiskit_convert_test.py
```diff
@@ -267,17 +269,14 @@ def test_tketpass() -> None:
     assert np.allclose(u1, u2)
 
 
-def test_tketautopass() -> None:
+@pytest.mark.skipif(skip_remote_tests, reason=REASON)
```
cqc-melf, line 272: should we add a no timeout mark to this test as well? Have you thought about that?

pytket/extensions/qiskit/backends/ibm.py
```diff
@@ -124,14 +122,16 @@ def _save_ibmq_auth(qiskit_config: Optional[QiskitConfig]) -> None:
...
+    except:
+        if token is not None:
+            IBMProvider.save_account(token, overwrite=True)
+            IBMProvider()
         else:
-            if token is not None:
-                IBMQ.save_account(token)
-            else:
-                raise NoIBMQAccountError()
+            raise NoIBMQAccountError()
```
cqc-melf, line 134: Should this give more details? (That we have not been able to load a token for example?)

pytket/extensions/qiskit/backends/ibm.py
```diff
@@ -124,14 +122,16 @@ def _save_ibmq_auth(qiskit_config: Optional[QiskitConfig]) -> None:
...
-    if not IBMQ.active_account():
-        if IBMQ.stored_account():
-            IBMQ.load_account()
+    if token is None and os.getenv("PYTKET_REMOTE_QISKIT_TOKEN") is not None:
+        token = os.getenv("PYTKET_REMOTE_QISKIT_TOKEN")
+    try:
+        IBMProvider()
+    except:
+        if token is not None:
+            IBMProvider.save_account(token, overwrite=True)
```
cqc-melf, line 131: Are there still cases where the overwrite is required?

pytket/extensions/qiskit/backends/ibm.py
CalMacCQ: The error message that will be given when this is raised will be the following 

```
No IBMQ credentials found on disk, store your account using qiskit, or using :py:meth:`pytket.extensions.qiskit.set_ibmq_config` first.
```

pytket/extensions/qiskit/backends/ibm.py
CalMacCQ: I'll leave this as it is for now and will make Yoshi's suggested change in a subsequent PR.

tests/qiskit_convert_test.py
CalMacCQ: I'll try without

tests/backend_test.py
```diff
@@ -197,14 +196,9 @@ def test_noise() -> None:
 
 
 @pytest.mark.skipif(skip_remote_tests, reason=REASON)
-def test_process_characterisation() -> None:
-    if not IBMQ.active_account():
-        IBMQ.load_account()
-
-    provider = IBMQ.providers(hub="ibm-q", group="open")[0]
-    back = provider.get_backend("ibmq_manila")
+def test_process_characterisation(manila_backend: IBMQBackend) -> None:
```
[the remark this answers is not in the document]
CalMacCQ: I will try this and see if it helps

tests/qiskit_convert_test.py
CalMacCQ: I will try without a timeout on this one as well.

tests/backend_test.py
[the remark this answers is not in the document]
CalMacCQ: Seems to pass :) I'm not sure if the timeout was the cause of the previous errors though

docs/intro.txt
```diff
@@ -60,35 +60,37 @@ Once you have created an account you can obtain an API token which you can use t
...
+To see which devices you can access you can use the ``available_devices`` method on the :py:class:`IBMQBackend` or :py:class:`IBMQEmulatorBackend`. Note that it is possible to pass optional ``instance`` and ``provider`` arguments to this method. This allows you to see which devices are accessible through your IBM hub.
 
 ::
 
     from pytket.extensions.qiskit import IBMQBackend
+    from qiskit_ibm_provider import IBMProvider
 
-    backend = IBMQBackend # Initialise backend for an IBM device
-    backend.available_devices(hub='your hub', group='your group', project='your project') 
+    ibm_provider = IBMProvider()
```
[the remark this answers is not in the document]
CalMacCQ: I guess its not essential. I was just showing how you can pass a provider as an optional argument to `IBMQBackend.available_devices()`

Do you think its better if I remove the line?

docs/intro.txt
```diff
@@ -60,35 +60,37 @@ Once you have created an account you can obtain an API token which you can use t
...
 locally without saving the token in pytket config:
 
 ::
 
-    from qiskit import IBMQ
+    from qiskit_ibm_provider import IBMProvider
     from qiskit_ibm_runtime import QiskitRuntimeService
 
-    IBMQ.save_account(token=ibm_token)
+    IBMProvider.save_account(token=ibm_token, overwrite=True)
```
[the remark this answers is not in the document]
CalMacCQ: I'll remove it from the docs for now. Maybe we should add separate documentation on why to do this once we make the change suggested by Yoshi.

cqc-melf (approved):

PR #126, opened 2023-06-22
Fix/available devices

CalMacCQ:
This PR fixes a bug with `IBMQBackend.available_devices()` introduced in https://github.com/CQCL/pytket-qiskit/pull/116

basically the version of with the instance string `IBMProvider(instance="ibm-q/open/main")` was always being overwritten by `IBMProvider()` due to the fact that an else statement was missing.

### DIFF
--- a/docs/intro.txt
+++ b/docs/intro.txt
@@ -88,9 +88,10 @@ To see which devices you can access you can use the ``available_devices`` method
     from pytket.extensions.qiskit import IBMQBackend
     from qiskit_ibm_provider import IBMProvider
 
-    ibm_provider = IBMProvider()
+    ibm_provider = IBMProvider(instance=f"{hub}/{group}/{project}")
     backend = IBMQBackend("ibmq_belem") # Initialise backend for an IBM device
-    backend.available_devices(instance=instance=f"{hub}/{group}/{project}", provider=ibm_provider) 
+    backendinfo_list = backend.available_devices(instance=f"{hub}/{group}/{project}", provider=ibm_provider) 
+    print([backend.device_name for backend in backendinfo_list])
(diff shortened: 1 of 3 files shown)


cqc-melf (commented):
Looks good!

The only thing we should add is the update to the changelog which is still missing the change from the project / hub arguments to the instance argument. Can you add a comment for that?

.github/workflows/docs/intro.txt
```diff
@@ -1,10 +0,0 @@
-pytket-extensions
```
cqc-melf, line 1: What happened to this file?

docs/intro.txt
```diff
@@ -73,6 +75,8 @@ If you are a member of an IBM hub then you can add this information to ``set_ibm
 Alternatively you can use the following qiskit commands to save your credentials
 locally without saving the token in pytket config:
 
+.. note:: If using pytket-qiskit 0.39.0 or older you will have to use the deprecated :py:meth:`IBMQ.save_account` instead of :py:meth:`IBMProvider.save_account` in the code below.
```
cqc-melf, line 78: Should we say that this is not recommended or so?

.github/workflows/docs/intro.txt
CalMacCQ: This was unused so I deleted it. I think its a legacy file from when all of the extensions were in the same repo.

docs/intro.txt
CalMacCQ: I think we should keep this as is for the next couple releases and maybe recommend an upgrade in the changelog?

cqc-melf (approved):

PR #128, opened 2023-06-22
Release/0.40.0

cqc-melf:


cqc-alec (approved):

pytket/extensions/qiskit/backends/ibm.py
```diff
@@ -124,19 +125,48 @@ def _save_ibmq_auth(qiskit_config: Optional[QiskitConfig]) -> None:
     token = None
     if qiskit_config is not None:
         token = qiskit_config.ibmq_api_token
-    if not IBMQ.active_account():
-        if IBMQ.stored_account():
-            IBMQ.load_account()
+    if token is None and os.getenv("PYTKET_REMOTE_QISKIT_TOKEN") is not None:
```
CalMacCQ, line 128: Do we need to keep this line in the released version?

docs/intro.txt
```diff
@@ -50,45 +50,53 @@ The :py:class:`AerBackend` also supports GPU simulation which can be configured
...
     from pytket.extensions.qiskit import IBMQBackend
+    from qiskit_ibm_provider import IBMProvider
 
-    backend = IBMQBackend # Initialise backend for an IBM device
-    backend.available_devices(hub='your hub', group='your group', project='your project') 
+    my_instance=f"{hub}/{group}/{project}"
+    ibm_provider = IBMProvider(instance=my_instance)
+    backend = IBMQBackend("ibmq_belem") # Initialise backend for an IBM device
+    backendinfo_list = backend.available_devices(instance=my_instance, provider=ibm_provider) 
+    print([backend.device_name for backend in backendinfo_list])
```
CalMacCQ, line 99: Do you both agree that the documentation here sufficently covers the changes the user will need to make when using 0.40.0?

pytket/extensions/qiskit/backends/ibm.py
CalMacCQ, line 128: I'm assuming we are keeping this workaround with the remote token in the released version and we have not found a better way to avoid this yet?

pytket/extensions/qiskit/backends/ibmq_emulator.py
```diff
@@ -83,19 +80,16 @@ def __init__(
...
             token=token,
         )
 
         self._service = QiskitRuntimeService(channel="ibm_quantum", token=token)
         self._session = Session(service=self._service, backend="ibmq_qasm_simulator")
 
         # Get noise model:
-        aer_sim = AerSimulator.from_backend(self._ibmq._backend)
-        self._noise_model = NoiseModel.from_backend(aer_sim)
+        self._noise_model = NoiseModel.from_backend(self._ibmq._backend)
```
CalMacCQ, line 92: Are we all satisfied that this is an okay way to get the `NoiseModel` for the `IBMQEmulatorBackend` instead of going via Aer?

tests/backend_test.py
```diff
@@ -1132,21 +1119,21 @@ def test_cloud_stabiliser(simulator_stabilizer_backend: IBMQBackend) -> None:
...
-    backend_info_list = IBMQBackend.available_devices(
-        hub="ibm-q", group="open", project="main"
-    )
+def test_available_devices(ibm_provider: IBMProvider) -> None:
+    backend_info_list = IBMQBackend.available_devices(instance="ibm-q/open/main")
     assert len(backend_info_list) > 0
 
-    provider = IBMQ.providers(hub="ibm-q", group="open")[0]
+    # Check consistency with pytket-qiskit and qiskit provider
+    assert len(backend_info_list) == len(ibm_provider.backends())
```
CalMacCQ, line 1127: This test ensures that the length of the list given by `IBMQBackend.available_devices()` is the same as the length of the list of devices in `IBMProvider.backends()` with the same `instance` string used in both. This doesn't check anything about the devices in the list only that the lengths match. 

Is this check thorough enough for testing purposes?

CalMacCQ (approved):

pytket/extensions/qiskit/backends/ibm.py
cqc-alec: I hadn't seen this change, why was it needed? Seems like the wrong place to be fixing a problem with our CI.

Issue #19, opened 2022-11-09
Eliminate warnings from test run


cqc-alec:
Some of these may be outside our control, in which case we can suppress them; others we should fix:
```
../env/lib/python3.10/site-packages/_pytest/fixtures.py:227
  /home/alec/r/pytket-qiskit/env/lib/python3.10/site-packages/_pytest/fixtures.py:227: PendingDeprecationWarning: The qiskit.Aer entry point will be deprecated in a future release and subsequently removed. Instead you should use this directly from the root of the qiskit-aer package.
    fixturemarker: Optional[FixtureFunctionMarker] = getattr(

tests/backend_test.py::test_operator
tests/backend_test.py::test_operator
tests/backend_test.py::test_operator
tests/backend_test.py::test_operator
tests/backend_test.py::test_expectation_bug
tests/backend_test.py::test_aer_placed_expectation
tests/backend_test.py::test_aer_placed_expectation
  /home/alec/r/pytket-qiskit/pytket/extensions/qiskit/backends/aer.py:710: PendingDeprecationWarning: The PauliTable class has been superseded by PauliList and is pending deprecation. This class will be deprecated in the future release and subsequently removed after that.
    return SparsePauliOp(PauliTable(table_array), coeffs)

tests/backend_test.py::test_operator
tests/backend_test.py::test_expectation_bug
tests/backend_test.py::test_aer_placed_expectation
tests/backend_test.py::test_aer_placed_expectation
  /home/alec/r/pytket-qiskit/env/lib/python3.10/site-packages/qiskit/quantum_info/operators/symplectic/sparse_pauli_op.py:131: ComplexWarning: Casting complex values to real discards the imaginary part
    self._coeffs = np.asarray((-1j) ** (phase - count_y) * coeffs, dtype=coeffs.dtype)

tests/qiskit_backend_test.py: 12 warnings
  /home/alec/r/pytket-qiskit/env/lib/python3.10/site-packages/qiskit_aer/noise/device/models.py:365: UserWarning: Device model returned an invalid T_2 relaxation time greater than the theoretical maximum value 2 * T_1 (94238.37625242842 > 2 * 2953.117923413739). Truncating to maximum value.
    warn("Device model returned an invalid T_2 relaxation time greater than"

tests/qiskit_backend_test.py::test_qiskit_counts
tests/qiskit_backend_test.py::test_qiskit_counts
[...]

PR #132, opened 2023-06-23
Eliminate warnings from test run via `pytest.ini`

robotAstray:
# Summary

This pull request aims to fix [issue 19](https://github.com/CQCL/pytket-qiskit/issues/19)

# Details & Comments

To fix [issue 19](https://github.com/CQCL/pytket-qiskit/issues/19) a simple modification has been made to `pytest.ini`. This modification ignores warnings when  running tests.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,6 +1,11 @@
 Changelog
 ~~~~~~~~~
 
+Unreleased
+----------
+
+* Fix conversion of qiskit `UnitaryGate` to and from pytket (up to 3 qubits).
+
 0.40.0 (June 2023)
 ------------------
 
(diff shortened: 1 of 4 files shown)


cqc-melf:
Hi @robotAstray 
Thank you for taking a look on the issue.

I think there might be a misunderstanding.
The idea was that each of the warnings should be investigates and the cause of it should be found. If it is possible to solve the cause of the warning that should be done as well.
In cases that the warnings can't be solved within pytket-qiskit they can just stay, they are not causing a failure of the test run.

robotAstray:
@cqc-alec I see, ok I will try to work on it!

CalMacCQ:
I think a lot of the test warnings are due to the fact there are still some deprecatred qiskit content in the tests

robotAstray:
@CalMacCQ @cqc-alec Good morning, I made some changes to my local repository and I will open a draft pull request shortly. I still have more changes to add.

Issue #134, opened 2023-06-28
Limited handling of `ControlledGate`(s) in `qiskit_to_tk`


CalMacCQ:
Following PR https://github.com/CQCL/pytket-qiskit/pull/49 qiskit `ContolledGate`s where the base gate is supported should be converted to `QContolBox`es in the converted circuit.

However the ability to do this is currently limited. Consider the following example with the CSGate.

```python
from qiskit import QuantumCircuit
from qiskit.circuit.library.standard_gates import CSGate
from pytket.extensions.qiskit import qiskit_to_tk

qc = QuantumCircuit(2)
qc.append(CSGate(), [0, 1])
tkc = qiskit_to_tk(qc)
```

This gives `KeyError: <class 'qiskit.circuit.library.standard_gates.s.CSGate'>`

However if instead I define a Controlled S gate manually it works fine.

```python
from qiskit.circuit.library.standard_gates import SGate

cs = SGate().control(1)
qc2 = QuantumCircuit(2)
qc2.append(cs , [0, 1])
tkc2 = qiskit_to_tk(qc2)
print(tkc2.get_commands())
```

This gives 
`[qif (q[0]) CircBox q[1];]`

So the `QControlBox` conversion works if the controlled gate is defined manually.

This is due to [this line](https://github.com/CQCL/pytket-qiskit/blob/develop/pytket/extensions/qiskit/qiskit_convert.py#L354). The `type(instr) == ControlledGate` check returns None for gates which inherit from `ControlledGate`.

I can fix this locally by using `if isinstance(instr, ControlledGate)` instead as this checks for inheritance. Howver this leads to test failures elsewhere as our current logic would convert a bunch of known qiskit gates to `QControlBox` es.

Issue #88, opened 2023-03-20
Some qiskit gates missing from the _known_qiskit_gate list


ferbetanzo:
After using `qiskit.circuit.random.random_circuit` for generating random qiskit circuits, if you try to transform those circuits to tket with `pytket.extensions.qiskit.qiskit_to_tk`, the process fails since there are some gates that are not recognised. The following is a list of the gates not recognised (beware that the list might not be complete):

```
qiskit.circuit.library.standard_gates.sx.CSXGate
qiskit.circuit.library.standard_gates.z.CCZGate
qiskit.circuit.library.standard_gates.rzx.RZXGate
qiskit.circuit.library.standard_gates.xx_plus_yy.XXPlusYYGate
qiskit.circuit.library.standard_gates.dcx.DCXGate
qiskit.circuit.library.standard_gates.xx_minus_yy.XXMinusYYGate
qiskit.circuit.library.standard_gates.s.CSdgGate
qiskit.circuit.library.standard_gates.x.RCCXGate
qiskit.circuit.library.standard_gates.s.CSGate
```

isobelhooper:
Other gates I found that were not recognised, when trying to do a similar thing with Qiskit's `random_circuit`:
* `qiskit.circuit.library.standard_gates.x.C3SXGate`
* `qiskit.circuit.library.standard_gates.x.RC3XGate`

CalMacCQ:
I would expect the controlled gates listed here to be handled in the converter after https://github.com/CQCL/pytket-qiskit/pull/50

Will take a look at this.

CalMacCQ:
Turns out the controlled gate coversion is more limited than expected https://github.com/CQCL/pytket-qiskit/issues/134.

Shouldn't be hard to fix this.

CalMacCQ:
FYI in 0.41.0 we now support CSX and CCZ gates directly.

There are some gates such as CS and C3SX where we support the base gate but don't have an OpType for the controlled gate itself. These will now be handled as `QControlBox`(es) post 0.41.0. This should have been the case before but I spotted a bug.

As for the other gates you point out, we may need to add new OpTypes if we want to support those ones.

PR #136, opened 2023-06-29
Fix conversion of controllled gates

CalMacCQ:
Adding the ability to handle the controlled gates mentioned in https://github.com/CQCL/pytket-qiskit/issues/88.
CCZ and CSX are both easily handled.

Also fixed the issue with Controlled gates not being handled correctly as `QControlBox`(es) in https://github.com/CQCL/pytket-qiskit/issues/134 .

### DIFF
--- a/pytket/extensions/qiskit/qiskit_convert.py
+++ b/pytket/extensions/qiskit/qiskit_convert.py
@@ -352,13 +352,18 @@ class CircuitBuilder:
                 pass
             self.add_xs(num_ctrl_qubits, ctrl_state, qargs)
             optype = None
-            if type(instr) == ControlledGate:
+            if isinstance(instr, ControlledGate):
                 if type(instr.base_gate) == qiskit_gates.RYGate:
                     optype = OpType.CnRy
                 elif type(instr.base_gate) == qiskit_gates.YGate:
                     optype = OpType.CnY
                 elif type(instr.base_gate) == qiskit_gates.ZGate:
                     optype = OpType.CnZ
+                elif (
+                    type(instr.base_gate) in _known_qiskit_gate
+                    and type(instr) in _known_qiskit_gate
+                ):
+                    optype = _known_qiskit_gate[type(instr)]
                 else:
                     if type(instr.base_gate) in _known_qiskit_gate:
                         optype = OpType.QControlBox  # QControlBox case handled below
--- a/pytket/extensions/qiskit/qiskit_convert.py
+++ b/pytket/extensions/qiskit/qiskit_convert.py
@@ -355,18 +355,12 @@ class CircuitBuilder:
             if isinstance(instr, ControlledGate):
                 if type(instr) in _known_qiskit_gate:
                     optype = _known_qiskit_gate[type(instr)]
-                    pass
                 elif type(instr.base_gate) == qiskit_gates.RYGate:
                     optype = OpType.CnRy
                 elif type(instr.base_gate) == qiskit_gates.YGate:
                     optype = OpType.CnY
                 elif type(instr.base_gate) == qiskit_gates.ZGate:
                     optype = OpType.CnZ
-                elif (
-                    type(instr.base_gate) in _known_qiskit_gate
-                    and type(instr) in _known_qiskit_gate
-                ):
-                    optype = _known_qiskit_gate[type(instr)]
                 else:
                     if type(instr.base_gate) in _known_qiskit_gate:
                         optype = OpType.QControlBox  # QControlBox case handled below
--- a/pytket/extensions/qiskit/qiskit_convert.py
+++ b/pytket/extensions/qiskit/qiskit_convert.py
@@ -353,7 +353,10 @@ class CircuitBuilder:
             self.add_xs(num_ctrl_qubits, ctrl_state, qargs)
             optype = None
             if isinstance(instr, ControlledGate):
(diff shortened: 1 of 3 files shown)


pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -350,13 +352,21 @@ def add_qiskit_data(self, data: "QuantumCircuitData") -> None:
                 pass
             self.add_xs(num_ctrl_qubits, ctrl_state, qargs)
             optype = None
-            if type(instr) == ControlledGate:
-                if type(instr.base_gate) == qiskit_gates.RYGate:
+            if isinstance(instr, ControlledGate):
+                if type(instr) in _known_qiskit_gate:
+                    optype = _known_qiskit_gate[type(instr)]
```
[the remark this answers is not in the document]
cqc-melf: it is used later on. Not sure why you are not happy with this?

tests/qiskit_convert_test.py
```diff
@@ -961,3 +961,42 @@ def test_unitary_gate() -> None:
...
     assert cmds[1].op.type == OpType.Unitary2qBox
     assert cmds[2].op.type == OpType.Unitary3qBox
+
+
+def test_ccz_conversion() -> None:
+    qc_ccz = QuantumCircuit(4)
+    qc_ccz.append(qiskit_gates.CCZGate(), [0, 1, 2])
+    qc_ccz.append(qiskit_gates.CCZGate(), [3, 1, 0])
+    tkc_ccz = qiskit_to_tk(qc_ccz)
+    assert tkc_ccz.n_gates_of_type(OpType.CnZ) == tkc_ccz.n_gates == 2
```
cqc-melf, line 971: Do you want to do any additional checks to find out if the qubits are set in the right way?

tests/qiskit_convert_test.py
```diff
@@ -961,3 +961,42 @@ def test_unitary_gate() -> None:
...
+    u1 = converted_tkc.get_unitary()
+    new_tkc_csx = Circuit(2)
+    new_tkc_csx.add_gate(OpType.CSX, [0, 1]).add_gate(OpType.CSX, [1, 0])
+    u2 = new_tkc_csx.get_unitary()
+    assert compare_unitaries(u1, u2)
+    converted_qc = tk_to_qiskit(new_tkc_csx)
+    assert converted_qc.count_ops()["csx"] == 2
+    qc_c3sx = QuantumCircuit(4)
+    qc_c3sx.append(qiskit_gates.C3SXGate(), [0, 1, 2, 3])
+    tkc_c3sx = qiskit_to_tk(qc_c3sx)
```
cqc-melf, line 991: Is it possible to convert this back to qiskit? Could we start with a tket circuit and convert it to qiskit and back and check if they are still identical?

CalMacCQ: It is possible in theory but we would have to add extra logic for that.

We could add a check to `tk_to_qiskit` that if it finds a `QControlBox` that checks the base gate of the QControl and then converts it to a qiskit gate with the right number of controls.

Problem is I think we'd have to somehow store the unsupported qiskit `ControlGate`s somewhere to do this.

tests/qiskit_convert_test.py
CalMacCQ: Do you mean check the unitary of the converted circuit to ensure that the conversion is not putting the control and target qubits in the wrong place?

pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -377,8 +383,8 @@ def add_qiskit_data(self, data: "QuantumCircuitData") -> None:
                 params = [param_to_tk(p) for p in instr.base_gate.params]
                 n_base_qubits = instr.base_gate.num_qubits
                 sub_circ = Circuit(n_base_qubits)
-                # use base gate name for the CircBox (shows in renderer)
-                sub_circ.name = instr.base_gate.name.capitalize()
+                # Use U for the name of the base gate in QControlBox
```
CalMacCQ, line 386: Do you think its better to call the base gate U or use the original name?

I guess I was wondering if it'd be confusing if a tket circuit has a gate that looks like CS in the renderer. The CS gate isn't directly supported and the gate would actually be a QControlBox.

tests/qiskit_convert_test.py
cqc-melf: Yes

tests/qiskit_convert_test.py
cqc-melf: Ha okay, in that case it might be better not to do that

pytket/extensions/qiskit/qiskit_convert.py
cqc-melf: I think u is fine, just do it as you like.

pytket/extensions/qiskit/qiskit_convert.py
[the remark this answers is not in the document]
CalMacCQ: The above comment is a bit outdated. 

Originally I thought there was still some duplication here but I think its okay now

tests/qiskit_convert_test.py
CalMacCQ: Done, let me know if theres anything else

cqc-melf (approved):

PR #139, opened 2023-06-30
Docs/cleanup and source links

CalMacCQ:
Updated copyright information.

Also added links to the source code using the sphinx.ext.viewcode as in https://github.com/CQCL/tket/pull/904

### DIFF
--- a/.github/workflows/docs/conf.py
+++ b/.github/workflows/docs/conf.py
@@ -3,7 +3,8 @@
 # Configuration file for the Sphinx documentation builder.
 # See https://www.sphinx-doc.org/en/master/usage/configuration.html
 
-author = "Cambridge Quantum Computing Ltd"
+copyright = "2023 Quantinuum"
+author = "Quantinuum"
 
 extensions = [
     "sphinx.ext.autodoc",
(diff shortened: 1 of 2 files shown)


.github/workflows/docs/conf.py
```diff
@@ -3,18 +3,18 @@
 # Configuration file for the Sphinx documentation builder.
 # See https://www.sphinx-doc.org/en/master/usage/configuration.html
 
-author = "Cambridge Quantum Computing Ltd"
+copyright = "2023 Quantinuum"
```
cqc-melf, line 6: Do we need to update this for the other extensions, too?

CalMacCQ: I guess this line is a bit redundant as in the old extensions docs build the copyright info is manually added to the page [here](https://github.com/CQCL/pytket-qiskit/blob/develop/.github/workflows/docs/build-docs#L72).

In the tket repo the copyright info is only specifed in conf.py

I have updated pytket-quantinuum already

I think it'd be worth updating the other extensions too but wouldn't say this is urgent.

CalMacCQ: I've removed the lines from `build-docs` so that the copyright info is only in `conf.py`.

PR #140, opened 2023-07-04
Improve Code Snippets: Add Import Statement and Clarify String Usage

spendierk:
These are suggested changes for a user who wants to simply copy the following code snippets and run them

suggest adding line 27:
`from pytket import Circuit`
since one gets _NameError: name 'Circuit' is not defined_ after running `pip install pytket-qiskit` and then wanting to try out this code snippet.

suggest adding `' '` around `ibm_token` in `set_ibmq_config(ibmq_api_token=ibm_token)`
because without it, one gets a _NameError: name '12345...' is not defined_ after simply inserting their copied API token i.e. `12345...` from IBM Q.

similar adding `' '` around `ibm_token` and adding `' '` around `hub`, `group`, `project`  in `set_ibmq_config(ibmq_api_token=ibm_token, instance=f"{hub}/{group}/{project}")` 
i.e one gets an error message without the `' '` for the entries inside the `instance` call.

and suggestions to include `' '` around `ibm_token` in
`IBMProvider.save_account(token=ibm_token)`
`QiskitRuntimeService.save_account(channel="ibm_quantum", token=ibm_token)`
and to include `' '` around  `hub`, `group`, `project`  in 
`my_instance=f"{hub}/{group}/{project}"`
follow from the above.

### DIFF
--- a/docs/intro.txt
+++ b/docs/intro.txt
@@ -83,8 +83,8 @@ locally without saving the token in pytket config:
     from qiskit_ibm_provider import IBMProvider
     from qiskit_ibm_runtime import QiskitRuntimeService
 
-    IBMProvider.save_account(token=ibm_token)
-    QiskitRuntimeService.save_account(channel="ibm_quantum", token=ibm_token)
+    IBMProvider.save_account(token="ibm_token")
+    QiskitRuntimeService.save_account(channel="ibm_quantum", token="ibm_token")
 
 To see which devices you can access you can use the ``available_devices`` method on the :py:class:`IBMQBackend` or :py:class:`IBMQEmulatorBackend`. Note that it is possible to pass optional ``instance`` and ``provider`` arguments to this method. This allows you to see which devices are accessible through your IBM hub.
 
--- a/docs/intro.txt
+++ b/docs/intro.txt
@@ -24,6 +24,7 @@ An example using the shots-based :py:class:`AerBackend` simulator is shown below
 ::
 
   from pytket.extensions.qiskit import AerBackend
+  from pytket import Circuit
 
   backend = AerBackend()
   circ = Circuit(2).H(0).CX(0, 1).measure_all()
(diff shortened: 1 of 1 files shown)


cqc-melf:
Thank you for the suggestion!

The documentation was written under the assumption that the values for the token, hub, group and project are stored in variables named that way, so you could just copy the code and work with it without changes.

I am not sure if it might be better to add the variable declaration than doing this, storing the token directly in the code / notebook might also be not the best idea.

If you would be happy with having a more detailed explanation I am happy to do that. If you think this would be better, we can do maybe a mix of both?

cqc-melf (commented):
Thank you for the suggestion!

The documentation was written under the assumption that the values for the token, hub, group and project are stored in variables named that way, so you could just copy the code and work with it without changes.

I am not sure if it might be better to add the variable declaration than doing this, storing the token directly in the code / notebook might also be not the best idea.

If you would be happy with having a more detailed explanation I am happy to do that. If you think this would be better, we can do maybe a mix of both?

spendierk:
Thank you for reviewing and explaining.

I understand the original context now (I did not catch this detail reading the doc). For users that are looking for a quick start. Including an example in the _Access and Credentials_ section might be beneficial then as you suggest, i.e:

`# Replace the placeholders with your actual values`
`ibm_token = '<your_ibm_token_here>'`
`hub = '<your_hub_here>'`
`group = '<your_group_here>'`
`project = '<your_project_here>'`

This addition would not alter the existing code too much, and clarify for quick start users.

cqc-melf:
> Thank you for reviewing and explaining.
> 
> I understand the original context now (I did not catch this detail reading the doc). For users that are looking for a quick start. Including an example in the _Access and Credentials_ section might be beneficial then as you suggest, i.e:
> 
> `# Replace the placeholders with your actual values` `ibm_token = '<your_ibm_token_here>'` `hub = '<your_hub_here>'` `group = '<your_group_here>'` `project = '<your_project_here>'`
> 
> This addition would not alter the existing code too much, and clarify for quick start users.

Yes, sounds good! I will add something like that

cqc-melf:
@spendierk I have created a new PR with the discussed changes:

https://github.com/CQCL/pytket-qiskit/pull/145

Feel free to take a look and leave some comments.

spendierk:
Looks good! Thank you

spendierk:
Figured out how to reopen. Just let me know what I need to do with this one

cqc-melf:
I think we can just close this one

PR #146, opened 2023-07-10
Simulate crosstalk

yao-cqc:


sjdilkes (changes requested):
A few minor changes, looking great.

pytket/extensions/qiskit/backends/crosstalk_model.py
```diff
@@ -0,0 +1,378 @@
...
+
+    :param zz_crosstalks: symmetric crosstalks between qubit pairs
+    :param single_q_phase_errors: dict specify the single qubit phase error on each qubit
+    :param two_q_induced_phase_errors: keys of dictionary specify the control and target qubit index,
+    while the values are tuples with the spectator qubit index and the amount of phase error to be applied.
+    :param non_markovian_noise: List storing the non-Markovian noise parameters.
+    Each tuple in the list contains the qubit index and the zx, zz noise parameters.
+    :param virtual_z: If True, then don't break any single qubit Z gate into unitary fractions,
+    instead add the full unitary.
+    :param N: hyperparameter N
```
daniel-mills-cqc, line 66: On a first pass it's not clear from this docstring how I should be choosing N. Is there a way to clarify this here?

pytket/extensions/qiskit/backends/aer.py
```diff
@@ -228,6 +233,17 @@ def process_circuits(
         if valid_check:
             self._check_all_circuits(circuits)
 
+        if hasattr(self, "crosstalk_params") and self.crosstalk_params is not None:
+            gate_times = get_gate_times_from_backendinfo(self.backend_info)
```
daniel-mills-cqc, line 237: I would be interested in discussing if this is the best way to obtain gate times. I would be okay with gate times being a property of `CrosstalkParams` to add user flexibility. I can't quite tell how easy it would be to build a purely hypothetical device with this set up (would I have to build a full set of characterisation information?).

pytket/extensions/qiskit/backends/crosstalk_model.py
```diff
@@ -0,0 +1,378 @@
...
+        circ: Circuit,
+        gate_times: Dict[Tuple[OpType, Tuple[Qubit, ...]], float],
+        ct_params: CrosstalkParams,
+    ) -> None:
+        """Construct a builder to generate noisy circuit
+        :param circ: the original circuit.
+        :type circ: `Circuit`
+        :param gate_times: python dict to store the gate time information.
+        :type gate_times: `Dict[Tuple[OpType, Tuple[Qubit, ...]], float]`
+        :param N: hyperparameter N
```
daniel-mills-cqc, line 100: Seems this is not actually a parameter? Maybe out of date?

pytket/extensions/qiskit/backends/crosstalk_model.py
```diff
@@ -0,0 +1,378 @@
...
+        :type N: float
+        :param ct_params: crosstalk parameters.
+        :type token: `CrosstalkParams`
```
daniel-mills-cqc, line 103:
```suggestion
        :type ct_params: `CrosstalkParams`
```

pytket/extensions/qiskit/backends/crosstalk_model.py
```diff
@@ -0,0 +1,378 @@
...
+                    gt = 1 / self.N
+                if gt is None:
+                    raise ValueError(
+                        f"No gate time for OpType {cmd.op.type} on qubits {cmd.qubits}"
+                    )
+                u = cmd.op.get_unitary()
+                if (self.N * gt) % 1 > 1e-10:
+                    raise ValueError(
+                        f"Command {cmd} cannot be factorised into equal slices"
+                    )
```
daniel-mills-cqc, line 211: I'm developing a better understanding of N as I read :P . So N is (seconds per slice)^{-1}? In particular the units of N are s^{-1}?

daniel-mills-cqc: This impose that the gates times should all have 'common factors'. I wonder how often that actually happens with device characterisation information? I think this makes it more important to make customisable gates times easy to input.

pytket/extensions/qiskit/backends/crosstalk_model.py
```diff
@@ -0,0 +1,378 @@
...
+            ZZ = zz / self.N
+            if abs(ZZ) > EPS:
+                noise_slice.append(
+                    NoiseGate(
+                        Command(Op.create(OpType.ZZPhase, ZZ), [q, two_level_q]),
+                        "non_markovian",
+                    )
+                )
+            if abs(ZX) > EPS:
+                RZX = Unitary2qBox(
```
sjdilkes, line 283: In the current setup, `Unitary2qBox` and `Unitary1qBox` in the Circuit slices require amplitude damping and phase damping, via a Qiskit NoiseModel`, while other gates added for other noise, such as the given non-Markovian noise, don't.

At a later date maybe we _should_ update our NoiseModel conversion to label things, but for now this constraint works for us. The only thing is this RZX is added as a `Unitary2qBox` so would have the same noise added to it, which we don't want.

There is a simple fix though, which is to do:
```
RZX = PauliExpBox([Pauli.Z, Pauli.X], ZX * math.pi/2)
```
Instead (though confirming the angle is correct, I may have written it wrong).

pytket/extensions/qiskit/backends/crosstalk_model.py
```diff
@@ -0,0 +1,378 @@
...
+    instead add the full unitary.
+    :param N: hyperparameter N
+    """
+
+    zz_crosstalks: Dict[Tuple[Qubit, Qubit], float]
+    single_q_phase_errors: Dict[Qubit, float]
+    two_q_induced_phase_errors: Dict[Tuple[Qubit, Qubit], Tuple[Qubit, float]]
+    non_markovian_noise: List[Tuple[Qubit, float, float]]
+    virtual_z: bool
+    N: float
```
sjdilkes, line 74: On further reflection, maybe this class should hold gate times and the amplitude damping and phase damping error rates?
So at least for the last two:
```
phase_damping_error: float
amplitude_damping_error: float
```

```
self.noise_model = qiskt_noise.NoiseModel(["unitary","rzz","rz","rzx"])

for

**Comment by daniel-mills-cqc** [comment]
Just had a quick look, and it seems great! Thanks a lot for this. Just to repeat some things I eluded to in my comments:
- I would like to discuss giving gate times as an input to CrosstalkParams for usability reasons.
- I would personally prefer not to allow the user to make the mistake of adding noise the the gates which are added for noisy purposes. That mistake is going to be very hard to notice and debug. Would it at least be possible to raise an exception if something like this happens.

File: pytket/extensions/qiskit/backends/crosstalk_model.py (hunk 7 above)
Comment (yao-cqc) [reply pytket/extensions/qiskit/backends/crosstalk_model.py:74]: how are these two errors going to be used?

Comment (sjdilkes) [reply pytket/extensions/qiskit/backends/crosstalk_model.py:74]: Actually it should be:
```
phase_damping_error: Dict[Qubit, float]
amplitude_damping_error: Dict[Qubit, float]
```

File: pytket/extensions/qiskit/backends/crosstalk_model.py [hunk 8]
```diff
@@ -0,0 +1,378 @@
...
+Slice = List[Instruction]
+EPS = 1e-9
+
+
+@dataclass
+class CrosstalkParams:
+    """
+    Stores various parameters for modelling crosstalk noise
+
+    :param zz_crosstalks: symmetric crosstalks between qubit pairs
```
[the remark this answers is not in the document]
CalMacCQ: I think I'm right in saying that the sphinx extension [sphinx-autodoc-typehints](https://github.com/tox-dev/sphinx-autodoc-typehints) infers the parameter types from the code so we don't need to update the type in multiple places.

Maybe we should be using this. We aren't using this at the moment.

yao-cqc:
Added customisable `gate_times`
Damping errors are added to error model.
Currently doesn't support classical gates and barriers. Circuit must only use default registers since qiskit NoiseModel only accepts integer qubits.

pytket/extensions/qiskit/backends/aer.py
```diff
@@ -440,6 +455,9 @@ def __init__(
          https://qiskit.org/documentation/stubs/qiskit.providers.aer.AerSimulator.html
          for available values. Defaults to "automatic".
         :type simulation_method: str
+        :param crosstalk_params: Apply crosstalk noise simulation to the circuits before
```
sjdilkes, line 458: I would updated the `crosstalk_params` description to explain that any other noise model will be overwritten.

sjdilkes (changes requested):
Just the one comment on making the `crosstalk_params` argument more descriptive: apart from that it looks great.

PR #155, opened 2023-08-08
Release/1.18

cqc-melf:


CalMacCQ:
Should we upgrade to qiskit 0.44.0 in the new release?

cqc-alec:
> Should we upgrade to qiskit 0.44.0 in the new release?

Yes I think so but I don't know how difficult that will be (#153 ). Can try updating the dependency on this branch but I suspect some other changes will be requires, in which case I'd probably wait as there aren't any other new features here.

cqc-melf:
I have tried that, but it is not working. I did not wanted to delay the release because of that. I would do that next week independent of this

cqc-melf:
And I think we still should do this release because of the changes to the documentation, which could be released

PR #157, opened 2023-08-09
remove nomidmeasure predicate

cqc-melf:


cqc-melf:
my understanding is that all back ends now support mid circuit measure

CalMacCQ:
Theres something I don't quite understand here.

These predicates have been part of the pytket-qiskit backends for a while.

Why is this only causing issues with the `compilation_example.py` example now? I don't think the predicates here have changed since the last release.

cqc-alec (changes requested):
This comes from the configuration reported by the qiskit backend, so I don't think we should override it.

CalMacCQ (approved):

cqc-melf:
From my understanding this should be fine, because all devices are now support mid circ measure, but if you want to keep this, it is fine with me.
Looks like this is not solving the error we had anyway :(

Issue #150, opened 2023-07-19
Improve error message for failed `qiskit_to_tk` conversion


CalMacCQ:
Currently if users try and convert a qiskit `QuantumCircuit` with an instruction that is not supported by `qiskit_to_tk` they get a rather unhelpful error message.

For example 

```python
from qiskit import QuantumCircuit
from qiskit.circuit.library import XXPlusYYGate
from pytket.extensions.qiskit import qiskit_to_tk

qc = QuantumCircuit(2)
qc.append(XXPlusYYGate(0.1), [0, 1]) # add unsupported gate

tkc = qiskit_to_tk(qc)
```

This gives the error

```
    optype = _known_qiskit_gate[type(instr)]
             ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
KeyError: <class 'qiskit.circuit.library.standard_gates.xx_plus_yy.XXPlusYYGate'
```

This error could be made more informative.

We should explain that the specific instruction is unsupported and maybe recommend using `QuantumCircuit.decompose()` to decompose higher level operations before conversion.

Of course we should still add new operations to the converter when we can.

CalMacCQ:
Addressed in https://github.com/CQCL/pytket-qiskit/pull/158.

Issue #43, opened 2022-12-19
Conversion from pytket to qiskit fails for circuits containing conditions on a partial register


cqc-alec:
```python
from pytket.circuit import Circuit
from pytket.extensions.qiskit import tk_to_qiskit

c = Circuit(1, 2)
c.X(0, condition_bits=[0], condition_value=1)

tk_to_qiskit(c)
```

Output:

```
NotImplementedError: OpenQASM conditions must be an entire register
```

The error message is misleading because we are not using OpenQASM for the conversion. It should be possible to convert such circuits to qiskit.

CalMacCQ:
Interestingly there are several error messages that are raised related to non qasm-compilant circuits. See [here](https://github.com/CQCL/pytket-qiskit/blob/develop/pytket/extensions/qiskit/qiskit_convert.py#L652) and the corresponding tests [here](https://github.com/CQCL/pytket-qiskit/blob/develop/tests/qiskit_convert_test.py#L403). This is indeed strange as we aren't using QASM for the conversion.

If I remove these lines of code it seems like qiskit circuits are converted incorrectly.

```python
from pytket.circuit import Circuit
from pytket.extensions.qiskit import tk_to_qiskit, qiskit_to_tk

tkc = Circuit(1, 2)
tkc.X(0, condition_bits=[0], condition_value=1)

qc = tk_to_qiskit(tkc)

tkc2 = qiskit_to_tk(qc)
```
`tkc` looks like this
<img width="210" alt="Screenshot 2023-08-10 at 11 15 46" src="https://github.com/CQCL/pytket-qiskit/assets/93673602/f6782f54-899e-4199-857d-d817fd62fdf6">

`tkc2` looks like this... (the X is conditoned on both of the bits).
<img width="214" alt="Screenshot 2023-08-10 at 11 15 58" src="https://github.com/CQCL/pytket-qiskit/assets/93673602/de13d5bc-e53b-42b6-badf-39bc9bca3440">

cqc-alec:
Closed by #158 .

cqc-alec:
Issue of inability to convert to qiskit remains.

PR #158, opened 2023-08-09
Improve error message for failed `qiskit_to_tk` conversion

CalMacCQ:
Small PR to imporve the error message for failed `qiskit_to_tk` conversion adressing https://github.com/CQCL/pytket-qiskit/issues/150

This snippet now gives the following error message for the unsupported gate

```python
from qiskit import QuantumCircuit
from qiskit.circuit.library import XXPlusYYGate
from pytket.extensions.qiskit import qiskit_to_tk

qc = QuantumCircuit(2)
qc.append(XXPlusYYGate(0.1), [0, 1]) # add unsupported gate

tkc = qiskit_to_tk(qc)
```

Screenshot 2023-08-11 at 11 02 48

Note: The two reverted commits correspond to an attempted fix of https://github.com/CQCL/pytket-qiskit/issues/43. However I realised this was slightly more involved than expected so its probably best adressed in a separate PR.

### DIFF
--- a/pytket/extensions/qiskit/qiskit_convert.py
+++ b/pytket/extensions/qiskit/qiskit_convert.py
@@ -379,7 +379,7 @@ class CircuitBuilder:
                 except KeyError:
                     raise NotImplementedError(
                         f"Conversion of qiskit's {instr.name} instruction is "
-                        + "currently unsupported by qiskit_to_tk. Perhaps try "
+                        + "currently unsupported by qiskit_to_tk. Consider "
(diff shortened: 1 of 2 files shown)


CalMacCQ:
Hmm maybe its better to add a custom error message here rather than `NotImplementedError`.

cqc-alec (approved):

cqc-alec:
> Hmm maybe its better to use custom error here rather than `NotImplementedError`.

`NotImplementedError` seems fine to me.

Issue #142, opened 2023-07-06
[AerUnitaryBackend] Call `transpile()` before `run()`


cqc-alec:
Seems necessary to handle some gates that act on large numbers of qubits, such as C5X: see https://github.com/Qiskit/qiskit-aer/issues/1855 .

doug-q:
reproducer:
```python
#!/usr/bin/env python3


from pytket.circuit import Circuit, CircBox, QControlBox
from pytket.extensions.qiskit import AerUnitaryBackend, AerStateBackend

backend = AerUnitaryBackend()
# backend = AerStateBackend() # using AerStateBackend does not raise an error

n_ancillas = 5 # using n_ancillas <=4 doees not raise an error
n_spins = 1
circ = Circuit(n_ancillas + n_spins)
trgt = Circuit(n_spins)
trgt.X(0)
# trgt.Y(0) # using Y instead of X does not raise an error

circ.add_qcontrolbox(
    QControlBox(CircBox(trgt), n_ancillas), list(range(n_ancillas + n_spins))
)

compiled_circ = backend.get_compiled_circuit(circ, optimisation_level = 0)
# using optimisation_level >= 1 does not raise an error
u = backend.run_circuit(compiled_circ)

```

PR #167, opened 2023-08-29
AerUnitaryBackend: call transpile before run (#142)

doug-q:
Fixes https://github.com/CQCL/pytket-qiskit/issues/142

### DIFF
--- a/tests/backend_test.py
+++ b/tests/backend_test.py
@@ -33,7 +33,7 @@ from qiskit_ibm_provider import IBMProvider  # type: ignore
 from qiskit_aer import Aer  # type: ignore
 from qiskit_ibm_provider.exceptions import IBMError  # type: ignore
 
-from pytket.circuit import Circuit, OpType, BasisOrder, Qubit, reg_eq, Unitary2qBox  # type: ignore
+from pytket.circuit import Circuit, OpType, BasisOrder, Qubit, reg_eq, Unitary2qBox, QControlBox, CircBox  # type: ignore
 from pytket.passes import CliffordSimp  # type: ignore
 from pytket.pauli import Pauli, QubitPauliString  # type: ignore
 from pytket.predicates import CompilationUnit, NoMidMeasurePredicate  # type: ignore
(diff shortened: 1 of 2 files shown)


tests/backend_test.py
```diff
@@ -1374,3 +1374,21 @@ def test_unitary_sim_gateset() -> None:
...
+    trgt = Circuit(n_spins)
+    trgt.X(0)
+
+    circ.add_qcontrolbox(
+        QControlBox(CircBox(trgt), n_ancillas), list(range(n_ancillas + n_spins))
+    )
+
+    compiled_circ = backend.get_compiled_circuit(circ, optimisation_level=0)
+    # using optimisation_level >= 1 does not raise an error
+    backend.run_circuit(compiled_circ)
```
cqc-alec, line 1394: Would be nice to have some sort of assertion in the test to check we got the expected result.

tests/qiskit_backend_test.py
```diff
@@ -107,15 +106,6 @@ def test_unitary() -> None:
         assert np.allclose(u, u2)
 
 
-def test_cancel() -> None:
```
cqc-alec, line 110: Why was this test deleted?

tests/backend_test.py
doug-q: I agree, do you have a suggestion? Otherwise I'll comment that we are testing that run_circuit doesn't throw

tests/qiskit_backend_test.py
doug-q: Awkward. Bad merge, although I don't understand how. Will revert.

tests/backend_test.py
cqc-alec: Something like:
```python
r = backend.run_circuit(compiled_circ)
u = r.get_unitary()
assert np.isclose(u[62:64, 62:64], np.asarray([[0., 1.].[1., 0.]]))
```
I think (may have got the syntax wrong, please check). I.e. check that the lower-right 2x2 submatrix of the unitary is the matrix of the X gate.

cqc-alec (approved):

PR #168, opened 2023-08-30
Fix/aer state backend gateset

CalMacCQ:
WIP

### DIFF
--- a/pytket/extensions/qiskit/backends/aer.py
+++ b/pytket/extensions/qiskit/backends/aer.py
@@ -96,6 +96,10 @@ def _tket_gate_set_from_qiskit_backend(
         gate_set.add(OpType.Unitary1qBox)
         gate_set.add(OpType.Unitary2qBox)
         gate_set.add(OpType.Unitary3qBox)
+
+    gate_set.add(OpType.Reset)
+    gate_set.add(OpType.Measure)
+    gate_set.add(OpType.Conditional)
     # special case mapping TK1 to U
     gate_set.add(OpType.TK1)
     return gate_set
--- a/pytket/extensions/qiskit/backends/aer.py
+++ b/pytket/extensions/qiskit/backends/aer.py
@@ -527,12 +527,12 @@ class AerStateBackend(_AerBaseBackend):
             version=__extension_version__,
             architecture=FullyConnected(n_qubits),
             gate_set=_tket_gate_set_from_qiskit_backend(self._qiskit_backend),
-            supports_midcircuit_measurement=True,  # is this correct?
+            supports_midcircuit_measurement=True,
+            supports_reset=True,
+            supports_fast_feedforward=True,
             misc={"characterisation": None},
         )
         self._required_predicates = [
-            NoClassicalControlPredicate(),
-            NoFastFeedforwardPredicate(),
             GateSetPredicate(self._backend_info.gate_set),
         ]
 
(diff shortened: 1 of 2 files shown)


pytket/extensions/qiskit/backends/aer.py
```diff
@@ -96,6 +96,10 @@ def _tket_gate_set_from_qiskit_backend(
         gate_set.add(OpType.Unitary1qBox)
         gate_set.add(OpType.Unitary2qBox)
         gate_set.add(OpType.Unitary3qBox)
+
+    gate_set.add(OpType.Reset)
+    gate_set.add(OpType.Measure)
+    gate_set.add(OpType.Conditional)
```
yao-cqc, line 102: I think not all qiskit simulators support these.

CalMacCQ: Annoyingly non-unitary ops such as measures and resets don't appear in the `basis_gates` field for the qiskit backend. I'll find some other way to check this.

Which simulators don't support these ops?

CalMacCQ: Given that we only support three wrapped aer simulators from pytket-qiskit (Unitary, statevector and standard shots based) I opted to just not add the additonal ops to the unitary simulator and added them to the other two.

tests/backend_test.py
```diff
@@ -1234,3 +1236,41 @@ def test_ecr_gate_compilation(ibm_sherbrooke_backend: IBMQBackend) -> None:
...
+    circ.H(2)
+    circ.Measure(2, 0)
+    circ.CZ(0, 1, condition_bits=[0], condition_value=1)
+    circ.add_gate(OpType.Reset, [2])
+    compiled_circ = sv_backend.get_compiled_circuit(circ)
+    assert sv_backend.valid_circuit(compiled_circ)
+    tket_statevector = sv_backend.run_circuit(compiled_circ).get_state()
+    qc = tk_to_qiskit(compiled_circ)
+    qiskit_statevector = _get_qiskit_statevector(qc)
+    assert compare_statevectors(tket_statevector, qiskit_statevector)
```
cqc-alec, line 1269: Was this circuit intentionally constructed to have a deterministic statevector?
I'm not sure how to interpret the result in general, but what would happen if there were a layer of hadamards at the start, for example? Would the statevector be probabilistic?

CalMacCQ: Yes this circuit is deterministic. Its taken from a paper and is part of a circuit that Lewis and I want to run for our IEEE demo.

Indeed if we inserted some Hadamards in front then the statevector test would sometimes fail as the circuit is non-deterministic.

I don't think this is a problem really as our `AerStateBackend` is a wrapper around qiskit's Aer statevector simulator. Given that they already support these non-unitary ops we are just ensuring that we support the same ops that qiskit can.

tests/backend_test.py
```diff
@@ -1235,3 +1237,43 @@ def test_ecr_gate_compilation(ibm_sherbrooke_backend: IBMQBackend) -> None:
...
+    return np.array(job.result().data()["statevector"].reverse_qargs().data)
+
+
+# https://github.com/CQCL/pytket-qiskit/issues/99
+def test_statevector_simulator_gateset() -> None:
+    sv_backend = AerStateBackend()
+    sv_supported_gates = sv_backend.backend_info.gate_set
+    assert OpType.Reset and OpType.Measure in sv_supported_gates
+    assert OpType.Conditional in sv_supported_gates
+    # This circuit is deterministic despite the non-unitary ops.
```
cqc-alec, line 1258: Need to be careful what we mean by "deterministic" here. Usually it means that the results of the final measurements are deterministic, but here we actually mean it has a well-defined (non-mixed) final state.

tests/backend_test.py
CalMacCQ: Good idea. I'll do that. Thanks Alec.

cqc-alec (approved):

PR #171, opened 2023-09-01
try import _TEMP_BIT_NAME from pytket.unit_id

trvto:


pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -73,7 +73,13 @@
     QControlBox,
     StatePreparationBox,
 )
-from pytket._tket.circuit import _TEMP_BIT_NAME  # type: ignore
+
+try:
+    from pytket.unit_id import _TEMP_BIT_NAME
```
cqc-alec, line 78: Do we need `#type: ignore` I wonder?

pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -73,7 +73,13 @@
...
+except:
```
cqc-alec, line 79:
```suggestion
except (ModuleNotFoundError, ImportError):
```

pytket/extensions/qiskit/qiskit_convert.py
trvto: maybe...

PR #187, opened 2023-10-13
Fix table.

cqc-alec:


docs/intro.txt
```diff
@@ -170,6 +170,7 @@ Every :py:class:`Backend` in pytket has its own ``default_compilation_pass`` met
      - self.rebase_pass [2]
      - `SynthesiseTket <https://cqcl.github.io/tket/pytket/api/passes.html#pytket.passes.SynthesiseTket>`_
    * -
+     -
      - self.rebase_pass [2]
```
CalMacCQ, line 174: I think when the table builds it shows the rebase pass applied after `RemoveRedundancies` (for optimisation level 1)

Shouldn't this be the other way round?

CalMacCQ: Ah okay, I must've got the ordering wrong initially then.

Thanks for fixing.

CalMacCQ (approved):

PR #193, opened 2023-10-30
Make dependabot check for updates to python package dependencies.

cqc-alec:
Let's see how this goes.


yao-cqc (approved):

cqc-melf (approved):
Do we want to add this to all extensions?

cqc-alec:
> Do we want to add this to all extensions?

Yes that was my intention.

PR #197, opened 2023-11-02
add seed list feature

cqc-melf:
Fixes the problem discussed in https://github.com/CQCL/pytket-qulacs/pull/44

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -9,7 +9,8 @@ unreleased
 * Update qiskit-ibm-runtime version to 0.13.0.
 * Update qiskit-aer version to 0.13.0.
 * Introduce dependency on qiskit-algorithms.
-* Add option for automatic incrementing seed on ``process_circuits()``.
+* Seed given to ``process_circuits()`` will be automaticly incremented
+  for the different circuit batches submitted.
 
 0.45.0 (October 2023)
 ---------------------
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -9,6 +9,7 @@ unreleased
 * Update qiskit-ibm-runtime version to 0.13.0.
 * Update qiskit-aer version to 0.13.0.
 * Introduce dependency on qiskit-algorithms.
+* Add possibility to submit list of seeds to backend in batch submission 
 
 0.45.0 (October 2023)
 ---------------------
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -9,7 +9,7 @@ unreleased
 * Update qiskit-ibm-runtime version to 0.13.0.
 * Update qiskit-aer version to 0.13.0.
 * Introduce dependency on qiskit-algorithms.
-* Add possibility to submit list of seeds to backend in batch submission 
+* Add option for automatic incrementing seed on ``process_circuits()``.
 
 0.45.0 (October 2023)
 ---------------------
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -9,7 +9,7 @@ unreleased
 * Update qiskit-ibm-runtime version to 0.13.0.
 * Update qiskit-aer version to 0.13.0.
 * Introduce dependency on qiskit-algorithms.
-* Seed given to ``process_circuits()`` will be automaticly incremented
+* Seed given to ``process_circuits()`` will be automatically incremented
   for the different circuit batches submitted.
 
 0.45.0 (October 2023)
--- a/pytket/extensions/qiskit/backends/aer.py
+++ b/pytket/extensions/qiskit/backends/aer.py
@@ -237,7 +237,7 @@ class _AerBaseBackend(Backend):
         circuits: Sequence[Circuit],
         n_shots: Union[None, int, Sequence[Optional[int]]] = None,
         valid_check: bool = True,
-        **kwargs: Union[bool, int, float, str, None],
+        **kwargs: KwargTypes,
     ) -> List[ResultHandle]:
         """
         See :py:meth:`pytket.backends.Backend.process_circuits`.
--- a/pytket/extensions/qiskit/backends/aer.py
+++ b/pytket/extensions/qiskit/backends/aer.py
@@ -237,8 +237,14 @@ class _AerBaseBackend(Backend):
         circuits: Sequence[Circuit],
         n_shots: Union[None, int, Sequence[Optional[int]]] = None,
         valid_check: bool = True,
-        **kwargs: Union[int, float, str, None, Sequence[Optional[int]]],
+        **kwargs: Union[bool, int, float, str, None],
     ) -> List[ResultHandle]:
+        """
+        See :py:meth:`pytket.backends.Backend.process_circuits`.
+        Supported kwargs: `seed`, `postprocess`, `seed_auto_increase`.
+        seed_auto_increase=True will automatically increase the seed by one for the
+        different batches when more than one circuit is submitted
+        """
         circuits = list(circuits)
         n_shots_list = Backend._get_n_shots_as_list(
             n_shots,
@@ -258,13 +264,12 @@ class _AerBaseBackend(Backend):
             circuits = noisy_circuits
 
         handle_list: List[Optional[ResultHandle]] = [None] * len(circuits)
-        circuit_batches, batch_order = _batch_circuits(
-            circuits, n_shots_list, kwargs.get("seed")
-        )
+        seed = kwargs.get("seed")
+        circuit_batches, batch_order = _batch_circuits(circuits, n_shots_list)
 
         replace_implicit_swaps = self.supports_state or self.supports_unitary
 
-        for (n_shots, seed, batch), indices in zip(circuit_batches, batch_order):
+        for (n_shots, batch), indices in zip(circuit_batches, batch_order):
             qcs = []
             for tkc in batch:
                 qc = tk_to_qiskit(tkc, replace_implicit_swaps)
@@ -284,6 +289,8 @@ class _AerBaseBackend(Backend):
                 seed_simulator=seed,
                 noise_model=self._noise_model,
             )
+            if kwargs.get("seed_auto_increase") and type(seed) is int:
+                seed += 1
             jobid = job.job_id()
             for i, ind in enumerate(indices):
                 handle = ResultHandle(jobid, i)
--- a/pytket/extensions/qiskit/backends/aer.py
+++ b/pytket/extensions/qiskit/backends/aer.py
@@ -241,9 +241,7 @@ class _AerBaseBackend(Backend):
     ) -> List[ResultHandle]:
         """
         See :py:meth:`pytket.backends.Backend.process_circuits`.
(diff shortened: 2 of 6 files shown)


cqc-alec:
Why is it necessary to provide a list of seeds, instead of just one seed which gets auto-incremented on each submission? This would give reproducibility and (assuming a half-decent PRNG) remove any bias. Is there a use case I'm missing?

cqc-melf:
> Why is it necessary to provide a list of seeds, instead of just one seed which gets auto-incremented on each submission? This would give reproducibility and (assuming a half-decent PRNG) remove any bias. Is there a use case I'm missing?

My understanding is that is this only taking affect when doing batch submissions. I was thinking of the case that you want to rerun your experiments for only a subset of your initial input. When doing auto incremented seeds this would not work.

cqc-alec (commented):
See https://github.com/CQCL/pytket-qiskit/pull/197#issuecomment-1794553905 .

cqc-melf:
> > My understanding is that is this only taking affect when doing batch submissions. I was thinking of the case that you want to rerun your experiments for only a subset of your initial input. When doing auto incremented seeds this would not work.
> 
> I think this is an unlikely use case, and quite confusing to use. If I understand the code, batching here is done by grouping together circuits with the same number of shots; so you would need one seed per _distinct value_ in your `n_shots` list, and when running a subset you would be very restricted in which subsets you could run (while still having reproducibility). All this would need to be explained in the docs. I think I'd rather keep things simple instead of adding a feature for which there's no clear need.

The list of shots contains one entry per circuit and in the same way I am expecting one entry in the seeds list per circuit.

If all the seeds are identical (or the user just gives one value like before) we can do the submissions of circuits batches using the qiskit-batch interface (this still works in the same way as before) (see https://github.com/CQCL/pytket-qiskit/pull/197/files#diff-8ccf4df5e4b2a5aa355a1a8a3086114f67945bdab047dc9f75ed25ed7dfd0850R78 )

If there are different seeds given, doing the creation of the qiskit-batches needs to be done differently: each batch now only contains exactly one circuit and the corresponding seed and shot number. (see https://github.com/CQCL/pytket-qiskit/pull/197/files#diff-8ccf4df5e4b2a5aa355a1a8a3086114f67945bdab047dc9f75ed25ed7dfd0850R92 )

If we want to do something like autoincrementing the seeds, we would need to break the qiskit-batch creation anyway. I have not found a way to have different seeds for the list of circuits in a qiskit batch.

So I think the user is free to choose any subset they would like and this would still work. (But if you think this is not a use case that is going to happen, I am happy to change to auto incrementing the seed)

cqc-alec:
> The list of shots contains one entry per circuit and in the same way I am expecting one entry in the seeds list per circuit.

Ah, this is indeed the case, but only because you have changed the logic of `_batch_circuits()`. It used to group circuits having the same number of shots together, but now it just puts one circuit in each batch (unless all the seeds are the same).

I assume this behaviour was in order to make most efficient use of the IBMQ interface so I think we should keep it.

cqc-melf:
> > The list of shots contains one entry per circuit and in the same way I am expecting one entry in the seeds list per circuit.
> 
> Ah, this is indeed the case, but only because you have changed the logic of `_batch_circuits()`. It used to group circuits having the same number of shots together, but now it just puts one circuit in each batch (unless all the seeds are the same).
> 
> I assume this behaviour was in order to make most efficient use of the IBMQ interface so I think we should keep it.

From my understanding the IBMQ interface only allows to set one seed for each qiskit-batch, so we would have to do equal seeds per qiskit batch. That would show the same issue as we had before, do we want to do that and would not be compatible with auto incrementing seeds as well?

cqc-alec:
> From my understanding the IBMQ interface only allows to set one seed for each qiskit-batch, so we would have to do equal seeds per qiskit batch. That would show the same issue as we had before, do we want to do that and would not be compatible with auto incrementing seeds as well?

What is the issue with one seed per qiskit batch? I would assume that the seed is used one to seed a global PRNG for the whole batch, not one PRNG per circuit.

cqc-alec:
If a user wants to have reproducibility at the sub-batch level, I think they need to design that into their workflow from the beginning: each call to `process_circuits()` should be reproducible, and therefore only needs a single seed.

cqc-melf:
I have removed the list feature and added a parameter for auto inc in [080f453](https://github.com/CQCL/pytket-qiskit/pull/197/commits/080f4535a763e62863531ab288b6cf29ab3a2ae3)

But I am not fully sure this is helpful like this, I would assume that this might still encourage people to not use batch submission when they want to reproduce results reliably.

pytket/extensions/qiskit/backends/aer.py
```diff
@@ -237,8 +237,14 @@ def process_circuits(
         circuits: Sequence[Circuit],
         n_shots: Union[None, int, Sequence[Optional[int]]] = None,
         valid_check: bool = True,
-        **kwargs: KwargTypes,
+        **kwargs: Union[bool, int, float, str, None],
     ) -> List[ResultHandle]:
+        """
+        See :py:meth:`pytket.backends.Backend.process_circuits`.
+        Supported kwargs: `seed`, `postprocess`, `seed_auto_increase`.
+        seed_auto_increase=True will automatically increase the seed by one for the
```
cqc-alec, line 245: I think this should be automatic (definitely the default behaviour, but I don't see a reasonable use case for _not_ incrementing the seed, which introduces biases in the results, so no need to add a new kwarg).

cqc-alec (changes requested):

cqc-alec:
> I have removed the list feature and added a parameter for auto inc in [080f453](https://github.com/CQCL/pytket-qiskit/pull/197/commits/080f4535a763e62863531ab288b6cf29ab3a2ae3)
> 
> But I am not fully sure this is helpful like this, I would assume that this might still encourage people to not use batch submission when they want to reproduce results reliably.

If incrementing the seed is automatic then every call to `process_circuits()` should be reproducible (if the same seed is used).

pytket/extensions/qiskit/backends/aer.py
cqc-melf: I did not wanted to do this by default because that would stop old runs from being reproducible, if that is not a concern I am happy to remove the parameter and always do the auto inc. The use case for the parameter I see is being compatible with reproducing old runs.

cqc-alec: At least they will still be reproducible if the same version numbers (of pytket-qiskit, qiskit, etc) are used.

docs/changelog.rst
```diff
@@ -9,6 +9,7 @@ unreleased
 * Update qiskit-ibm-runtime version to 0.13.0.
 * Update qiskit-aer version to 0.13.0.
 * Introduce dependency on qiskit-algorithms.
+* Add option for automatic incrementing seed on ``process_circuits()``.
```
cqc-alec, line 12: Update description in changelog.

pytket/extensions/qiskit/backends/aer.py
cqc-alec, line 245: Remove.

pytket/extensions/qiskit/backends/aer.py
```diff
@@ -275,14 +282,15 @@ def process_circuits(
...
-            seed = cast(Optional[int], kwargs.get("seed"))
             job = self._qiskit_backend.run(
                 qcs,
                 shots=n_shots,
                 memory=self._memory,
                 seed_simulator=seed,
                 noise_model=self._noise_model,
             )
+            if kwargs.get("seed_auto_increase") and type(seed) is int:
+                seed += 1
```
cqc-alec, line 293:
```suggestion
            seed += 1
```

pytket/extensions/qiskit/backends/ibmq_emulator.py
```diff
@@ -125,11 +125,13 @@ def process_circuits(
...
         valid_check: bool = True,
-        **kwargs: KwargTypes,
+        **kwargs: Union[bool, int, float, str, None],
     ) -> List[ResultHandle]:
         """
         See :py:meth:`pytket.backends.Backend.process_circuits`.
-        Supported kwargs: `seed`, `postprocess`.
+        Supported kwargs: `seed`, `postprocess`, `seed_auto_increase`.
+        seed_auto_increase=True will automatically increase the seed by one for the
+        different batches when more than one circuit is submitted
```
cqc-alec, line 134: Remove.

pytket/extensions/qiskit/backends/ibmq_emulator.py
```diff
@@ -173,7 +176,9 @@ def process_circuits(
                 options.resilience_level = 0
                 options.execution.shots = n_shots
                 options.simulator.noise_model = self._noise_model
-                options.seed_simulator = kwargs.get("seed")
+                options.seed_simulator = seed
+                if kwargs.get("seed_auto_increase") and type(seed) is int:
+                    seed += 1
```
cqc-alec, line 181:
```suggestion
                seed += 1
```

cqc-alec (changes requested):

pytket/extensions/qiskit/backends/ibmq_emulator.py
cqc-melf: Done in [85489ef](https://github.com/CQCL/pytket-qiskit/pull/197/commits/85489ef983300e4739976ba68048e55e5ceb932a)

pytket/extensions/qiskit/backends/ibmq_emulator.py
cqc-melf: Done in [85489ef](https://github.com/CQCL/pytket-qiskit/pull/197/commits/85489ef983300e4739976ba68048e55e5ceb932a)

pytket/extensions/qiskit/backends/aer.py
cqc-melf: Done in [85489ef](https://github.com/CQCL/pytket-qiskit/pull/197/commits/85489ef983300e4739976ba68048e55e5ceb932a)

pytket/extensions/qiskit/backends/aer.py
cqc-melf: Done in [85489ef](https://github.com/CQCL/pytket-qiskit/pull/197/commits/85489ef983300e4739976ba68048e55e5ceb932a)

docs/changelog.rst
cqc-melf: Done in [85489ef](https://github.com/CQCL/pytket-qiskit/pull/197/commits/85489ef983300e4739976ba68048e55e5ceb932a)

pytket/extensions/qiskit/backends/ibmq_emulator.py
cqc-melf: I needed to keep the type check for mypy. (and to make sure I am not incrementing when the seed is set to none)

cqc-alec (approved):

PR #219, opened 2023-12-06
Add classes for crosstalk simulation to API docs

CalMacCQ:


docs/api.rst
```diff
@@ -2,35 +2,44 @@ API documentation
...
 
 .. automodule:: pytket.extensions.qiskit
     :members: qiskit_to_tk, tk_to_qiskit, process_characterisation
 
 .. automodule:: pytket.extensions.qiskit.tket_backend
+    :show-inheritance:
     :members:
     :special-members: __init__
 
+.. automodule:: pytket.extensions.qiskit.backends.crosstalk_model
```
CalMacCQ, line 40: I've added the three classes `CrosstalkParams`, `NoisyCircuitBuilder` and `FractionalUnitary`. 

These are the only three classes which have tests so I assumed they were the only user facing classes. Is that correct?

pytket/extensions/qiskit/backends/crosstalk_model.py
```diff
@@ -74,28 +75,20 @@ class CrosstalkParams:
     Stores various parameters for modelling crosstalk noise
```
CalMacCQ, line 76: These types will be automatically inferred by the sphinx-autodoc-typehints extension. We won't have to write them down in multiple places.

yao-cqc (approved):
Looks good!

docs/api.rst
CalMacCQ: Update: Only `CrosstalkParams` is user facing

PR #287, opened 2024-03-06
fix brisbane not working with ibmq_qasm_simulator anymore

cqc-melf:
# Description

All devices available in the open plan have more then the 32 qubits that are supported by the ibmq_qasm_simulator. So we need to replace them. Not sure what the best strategy is... 

# Related issues

# Checklist

### DIFF
--- a/tests/backend_test.py
+++ b/tests/backend_test.py
@@ -856,6 +856,8 @@ def test_ibmq_emulator(
 
             c_cop_2 = c.copy()
             c_cop_2 = b_aer.get_compiled_circuit(c_cop_2, ol)
+            if ol == 0:
+                assert not all(pred.verify(c_cop_2) for pred in b.required_predicates)
 
         circ = Circuit(2, 2).H(0).CX(0, 1).measure_all()
         copy_circ = circ.copy()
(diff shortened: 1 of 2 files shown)


tests/backend_test.py
```diff
@@ -857,8 +857,8 @@ def test_ibmq_emulator(
 
             c_cop_2 = c.copy()
             c_cop_2 = b_aer.get_compiled_circuit(c_cop_2, ol)
-            if ol == 0:
-                assert not all(pred.verify(c_cop_2) for pred in b.required_predicates)
+            # if ol == 0:
```
cqc-melf, line 860: The circuit is already fully supported by the simulator without compilation, so this test fails with the change

cqc-melf:
Ready for review.
I know that the tests are not working, but that is a different issue I am still investigating.

cqc-alec:
What exactly is the issue with `brisbane_emulator_backend`? I see that it is used in several other tests, such as `test_nshots()`...

cqc-melf:
This difference is that this testcase is the only test using the backend and compiling the circuit with one of the default compilation passes. All other testcases are running uncompiled circuits.

cqc-melf:
When running this with the `brisbane_emulator_backend` the circuit has 51 qubits (only 32 allowed), and the circuit contains `Ecr` gates, which is not allowed, too.

cqc-alec:
> When running this with the `brisbane_emulator_backend` the circuit has 51 qubits (only 32 allowed), and the circuit contains `Ecr` gates, which is not allowed, too.

These both seem to indicate a bug in the compilation:
* Why is a 3-qubit circuit being compiled to a 51-qubit circuit?
* Why are we rebasing to ECR when that isn't supported by the backend?

cqc-melf:
The IBMQEmulatorBackend is using the gateset, architecture, noisemodel and compilation passes of the real device, in our case the IBMQBackend. The IBMQEmulatorBackend is then always running the circuit on the simulator without additional compilation or checks. We could add that, but not sure how much would be left from the original device after that.

cqc-melf:
@cqc-alec this is currently not showing up on develop, but I am not sure why. I would keep this open and wait a few days?

PR #315, opened 2024-04-05
add testcase for symbolic conversion

cqc-melf:
# Description

Adding tests for https://github.com/CQCL/pytket-qiskit/pull/314 

# Related issues

# Checklist

### DIFF
--- a/tests/qiskit_convert_test.py
+++ b/tests/qiskit_convert_test.py
@@ -35,6 +35,9 @@ from qiskit.transpiler.passes import BasisTranslator  # type: ignore
 from qiskit.circuit.equivalence_library import StandardEquivalenceLibrary  # type: ignore
 from qiskit_ibm_runtime.fake_provider import FakeGuadalupe  # type: ignore
 from qiskit.circuit.parameterexpression import ParameterExpression  # type: ignore
+from pytket.extensions.qiskit import qiskit_to_tk, tk_to_qiskit
+from qiskit.circuit.library import TwoLocal
+from qiskit import transpile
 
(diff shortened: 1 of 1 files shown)


tests/qiskit_convert_test.py
```diff
@@ -1091,3 +1094,24 @@ def test_RealAmplitudes_numeric_params() -> None:
     unitary2 = tkc2.get_unitary()
     assert compare_unitaries(qiskit_unitary, unitary1)
     assert compare_unitaries(unitary1, unitary2)
+
+
+# https://github.com/CQCL/pytket-qiskit/issues/256
+def test_symbolic_param_conv() -> None:
```
CalMacCQ, line 1100: I note that this test does not have an assert statement or anything of that kind.

Is the idea just to ensure that this code executes successfully and doesn't give the 
```
TypeError: EvalfMixin.evalf() got an unexpected keyword argument 'real'
```

As before?

CalMacCQ (commented):
One clarification needed. Also can you link to the orignal PR in the desciption please?

tests/qiskit_convert_test.py
cqc-melf: Yes that was my idea, I was thinking about checking stuff in the final circuit, but I would expect the way qiskit is setting up the circuit or the compilation works might change without generating any issues here, so I decided not to do any additional checks.

cqc-melf:
> One clarification needed. Also can you link to the orignal PR in the desciption please?

Done

CalMacCQ (approved):

PR #316, opened 2024-04-05
Feature/add qiskit api options

irfankhan10:
# Description

Pass `Qiskit `API Sampler options to the Sampler instance via the `IBMQBackend` constructor or via the instance method `process_circuit` (`process_circuits`)

# Checklist

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -5,8 +5,8 @@ Unreleased
 ----------
 
 * User can pass an `Options` instance (from `qiskit-ibm-runtime`)
-via a keyword argument to both an `IBMQBackend` constructor and 
-an instance method `IBMQBackend.process_circuits`.
+  via a keyword argument to both an `IBMQBackend` constructor and 
+  an instance method `IBMQBackend.process_circuits`.
 
 0.53.0 (April 2024)
 -------------------
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,6 +1,13 @@
 Changelog
 ~~~~~~~~~
 
+Unreleased
+----------
+
+* User can pass an `Options` instance (from `qiskit-ibm-runtime`)
+via a keyword argument to both an `IBMQBackend` constructor and 
+an instance method `IBMQBackend.process_circuits`.
+
 0.53.0 (April 2024)
 -------------------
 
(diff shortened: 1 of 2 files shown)


pytket/extensions/qiskit/backends/ibm.py
```diff
@@ -167,6 +167,10 @@ class IBMQBackend(Backend):
     :type provider: Optional[IBMProvider]
     :param token: Authentication token to use the `QiskitRuntimeService`.
     :type token: Optional[str]
+    :param options: A dictionary to specify qiskit API options,
```
cqc-alec, line 170: If these are the only supported options, I think it makes sense to have them as distinct kwargs, rather than having to pass them in a dictionary. Alternatively, why not just allow the user to pass an arbitrary qiskit `Options` object here?

irfankhan10: OK. Happy to make this change.

pytket/extensions/qiskit/backends/ibm.py
```diff
@@ -519,6 +540,11 @@ def process_circuits(
         postprocess = kwargs.get("postprocess", False)
         simplify_initial = kwargs.get("simplify_initial", False)
 
+        options: Options = kwargs.get("options")
+        if options is None:
+            options = self._sampler_options
+        options.execution.shots = n_shots
```
cqc-alec, line 546: The test is failing because `n_shots` here is different from `n_shots` below (where code was deleted). I think `options.execution_shots` needs to be set further down (when submitting an individual circuit).

irfankhan10: I have moved this lower down. The n_shots property is now set per chunk of a batch.

irfankhan10:
Hi @cqc-alec and @yao-cqc, everything passes here now.

cqc-alec (commented):
Thanks, please could you also add a note to the changelog (`docs/changelog.txt`), under a new "Unreleased" heading at the top of the file?

irfankhan10:
Added changelog as requested

docs/changelog.rst
```diff
@@ -1,6 +1,13 @@
 Changelog
 ~~~~~~~~~
 
+Unreleased
+----------
+
+* User can pass an `Options` instance (from `qiskit-ibm-runtime`)
+via a keyword argument to both an `IBMQBackend` constructor and 
+an instance method `IBMQBackend.process_circuits`.
```
cqc-alec, line 9:
```suggestion
* User can pass an `Options` instance (from `qiskit-ibm-runtime`)
  via a keyword argument to both an `IBMQBackend` constructor and 
  an instance method `IBMQBackend.process_circuits`.
```

cqc-melf (approved):

PR #335, opened 2024-05-24
Handle PrimitiveResult as result type.

### LINKED ISSUE #334: Job results from IBMQ devices are not (always) `SamplerResult`
The code currently assumes they are and errors out if not: https://github.com/CQCL/pytket-qiskit/blob/1d027ebc5c08e6aaa5265b0c7f2c6920c73f1187/pytket/extensions/qiskit/backends/ibm.py#L617

The actual type is `PrimitiveResult`.

(This issue does not affect the latest released version, but we should fix before the next release.)

cqc-alec:
# Description

We were assuming a `SamplerResult`. We still handle that case, though I don't know if it ever occurs.

# Related issues

Fixes #334 .

# Checklist

### DIFF
--- a/pytket/extensions/qiskit/backends/ibm.py
+++ b/pytket/extensions/qiskit/backends/ibm.py
@@ -32,7 +32,9 @@ from typing import (
 )
 from warnings import warn
 
-from qiskit.primitives import SamplerResult  # type: ignore
+import numpy as np
+
+from qiskit.primitives import PrimitiveResult, SamplerResult  # type: ignore
 
 
 # RuntimeJob has no queue_position attribute, which is referenced
@@ -136,6 +138,12 @@ def _get_primitive_gates(gateset: Set[OpType]) -> Set[OpType]:
         return {OpType.X, OpType.SX, OpType.Rz, OpType.ECR}
     else:
         return gateset
+
+
+def _int_from_readout(readout: np.ndarray) -> int:
+    # Weird mixture of big- and little-endian here.
+    n_bytes = len(readout)
+    return sum(int(x) << (8 * (n_bytes - 1 - i)) for i, x in enumerate(readout))
 
 
 class IBMQBackend(Backend):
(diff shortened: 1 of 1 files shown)


pytket/extensions/qiskit/backends/ibm.py
```diff
@@ -614,11 +622,19 @@ def get_result(self, handle: ResultHandle, **kwargs: KwargTypes) -> BackendResul
                         sleep(10)
 
                 res = job.result(timeout=kwargs.get("timeout", None))
-            assert isinstance(res, SamplerResult)
-            for circ_index, (r, d) in enumerate(zip(res.quasi_dists, res.metadata)):
-                self._ibm_res_cache[(jobid, circ_index)] = Counter(
-                    {n: int(0.5 + d["shots"] * p) for n, p in r.items()}
-                )
+            if isinstance(res, SamplerResult):
+                # TODO Is this code still reachable?
```
sjdilkes, line 626: Do we need to address this TODO here?

sjdilkes (changes requested):
LGTM, just want to check the TODO doesn't need to be addressed.

pytket/extensions/qiskit/backends/ibm.py
cqc-alec: I think it can wait; I will raise a new issue.

cqc-alec: #336

sjdilkes (approved):

pytket/extensions/qiskit/backends/ibm.py
```diff
@@ -614,11 +622,19 @@ def get_result(self, handle: ResultHandle, **kwargs: KwargTypes) -> BackendResul
...
+                for circ_index, (r, d) in enumerate(zip(res.quasi_dists, res.metadata)):
+                    self._ibm_res_cache[(jobid, circ_index)] = Counter(
+                        {n: int(0.5 + d["shots"] * p) for n, p in r.items()}
+                    )
+            else:
+                assert isinstance(res, PrimitiveResult)
+                for circ_index, pub_result in enumerate(res._pub_results):
+                    readouts = pub_result.data.c.array
```
mwpb, line 634: Is there something in the set up that allows us to assume that the register is called `c`?

mwpb: For me it is called `readout` for instance.

mwpb: A quick question: does it make sense to have multiple classical registers in a BackendResult? Or would we end up 'concatenating' them into one? If so then something like the following might prove useful?
```
counts = []
for result in res._pub_results:
    result = SamplerPubResult(result.data)
    counts.append(result.join_data().get_counts())
```
It gives a dictionary of the form `{'00': 7, '11': 3}` for instance. Hopefully that is reasonably easy to convert into a `BackendResult`? Alternatively I suppose we could use the shots directly but that would be a lot large object I think.

cqc-alec: Yes it does make sense, we just pack them in lexicographic order I think. Thanks for the suggestion -- I'll try it out.

mwpb: (Note that I've also ended up with a list of BackendResults, which I'm not quite sure if I should be merging or not. I think they can correspond to completely different circuits.)

PR #340, opened 2024-05-29
Handle PrimitiveResult as result type

### LINKED ISSUE #334: Job results from IBMQ devices are not (always) `SamplerResult`

cqc-alec:
# Description

We were assuming a `SamplerResult`. We still handle that case, though I don't know if it ever occurs.

This PR supersedes #335 .

# Related issues

Fixes #334 .

# Checklist

### DIFF
--- a/pytket/extensions/qiskit/backends/ibm.py
+++ b/pytket/extensions/qiskit/backends/ibm.py
@@ -50,7 +50,7 @@ from qiskit_ibm_runtime import (  # type: ignore
     RuntimeJob,
 )
 
-from pytket.circuit import Circuit, OpType
+from pytket.circuit import Bit, Circuit, OpType
 from pytket.backends import Backend, CircuitNotRunError, CircuitStatus, ResultHandle
 from pytket.backends.backendinfo import BackendInfo
 from pytket.backends.backendresult import BackendResult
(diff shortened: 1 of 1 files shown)


pytket/extensions/qiskit/backends/ibm.py
```diff
@@ -525,11 +533,14 @@ def process_circuits(
 
                 qcs, ppcirc_strs = [], []
                 for tkc in batch_chunk:
+                    tkc1 = tkc.copy()
+                    # Flatten bits to default register in lexicographic order:
```
cqc-alec, line 537: This solves the issue raised in [this comment](https://github.com/CQCL/pytket-qiskit/pull/335#discussion_r1613566199). There is still an issue with circuits having non-default classical register names, but this is an independent pre-existing issue and is not addressed in this PR.

mwpb: Just to make sure I understand: this does actually all work if `pytket-qiskit` is used to submit and retrieve the circuit no matter what the classical registers are named. Of course if one separately submits a circuit with non-default classical register names and tries to retrieve it with `pytket-qiskit` then there is an error... but that is indeed a different issue!

mwpb (approved):
Have tested this by hitting IBM using this local branch and everything seems to work well for me! Did have one question but is for my understanding and doesn't affect the PR. Feel free to merge or wait for others more familiar with the code to take a look also.

pytket/extensions/qiskit/backends/ibm.py
cqc-alec: Yes, it works _unless_ you then want to call `get_counts()` on the result with specific bits, using the `cbits` parameter to `get_counts`. In that case it will throw an error -- but that was already the case (separate issue).

mwpb: Thanks that makes sense!

PR #349, opened 2024-06-14
update dependencies

cqc-melf:
# Description

Please summarise the changes.

# Related issues

# Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,10 +44,10 @@ setup(
     packages=find_namespace_packages(include=["pytket.*"]),
     include_package_data=True,
     install_requires=[
-        "pytket ~= 1.27",
-        "qiskit ~= 1.0",
+        "pytket ~= 1.28",
+        "qiskit ~= 1.1",
         "qiskit-algorithms ~= 0.3.0",
-        "qiskit-ibm-runtime ~= 0.23.0",
+        "qiskit-ibm-runtime ~= 0.24.0",
         "qiskit-aer ~= 0.14.0",
         "numpy",
     ],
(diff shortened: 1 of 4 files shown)


cqc-melf:
I am not fully sure is the change to string is fully intentional, the old enum is still used in some parts?

setup.py
```diff
@@ -44,10 +44,10 @@
     packages=find_namespace_packages(include=["pytket.*"]),
     include_package_data=True,
     install_requires=[
-        "pytket ~= 1.27",
-        "qiskit ~= 1.0",
+        "pytket ~= 1.28",
```
cqc-alec, line 47: Can we go with the versions in #354 which are more recent?

cqc-melf: Yes we can

cqc-alec (approved):

setup.py
```diff
@@ -44,11 +44,11 @@
...
-        "pytket ~= 1.27",
-        "qiskit ~= 1.0",
+        "pytket ~= 1.29",
+        "qiskit ~= 1.1",
         "qiskit-algorithms ~= 0.3.0",
-        "qiskit-ibm-runtime ~= 0.23.0",
-        "qiskit-aer ~= 0.14.0",
+        "qiskit-ibm-runtime ~= 0.24.1",
+        "qiskit-aer ~= 0.14.2",
         "numpy",
```
cqc-alec, line 52: Ideally we'd have "< 2" here but that's already implied by pytket's requirements, so no need.

cqc-melf: I have added that. We might update pytket in the future, but we probably will face additional changes here.

cqc-alec (approved):

PR #358, opened 2024-06-24
Remove upper caps on dependency versions.

### LINKED ISSUE #351: Question regarding version requirements
Hi there, I have noticed that the qiskit and tket version requirements are in the format of 

```     
"pytket ~= 1.x",
"qiskit ~= 1.y",
```

This has the effect, that for each new release of those two packages of version `1.x.y` are included, e.g., when pytket was upgrade from `1.28.0 to 1.29.0`, the requirement of `pytket ~= 1.27` automatically included it (since 1.29.0 satisfies ~= 1.27). Would it possible to switch to

```     
"pytket ~= 1.x.0",
"qiskit ~= 1.y.0",
```
by including the patch version as well?

We are heavily using this package as a dependency and this versioning scheme has led to some unexpected changed behavior and makes versioning not so easy for us.

cqc-alec:
See discussion in #351 .

Closes #351 .

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,12 +44,12 @@ setup(
     packages=find_namespace_packages(include=["pytket.*"]),
     include_package_data=True,
     install_requires=[
-        "pytket ~= 1.29",
-        "qiskit ~= 1.1",
-        "qiskit-algorithms ~= 0.3.0",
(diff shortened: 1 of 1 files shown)


cqc-melf:
I think this should be explained in the changelog?

cqc-melf (approved):

PR #359, opened 2024-06-24
Update ibm.py to enable execution on IBM QPUs

salmma:
`status` has no attribute `name` for IBMQ QPUs (e.g. ibm_brisbane) such that an `AttributeError` occurs when executing on e.g. ibm_brisbane. 

_Note:_  This is not a problem for the AerSimulator, such that the tests are still working.


cqc-alec (approved):
Thanks!

PR #380, opened 2024-08-22
feat: add density matrix simulation via qiskit Aer

### LINKED ISSUE #231: Support density matrix simulation

CalMacCQ:
# Description

Adding a new `AerDensityMatrixBackend` class. Unlike other density matrix simulators available through the pytket extensions this emulator should support an optional `NoiseModel` in the constructor.

There's a decent amount of duplication from the `AerBackend` implementation for `BackendInfo`. Maybe there's a smarter and less error prone way to do it. 

# Related issues
closes #231 

# Checklist

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,9 +1,13 @@
 Changelog
 ~~~~~~~~~
 
+.. currentmodule:: pytket.extensions.qiskit
+
+
 Unreleased
 ----------
 
+* Added :py:class:`AerDensityMatrixBackend` simulator. This simulator has the option to support a :py:class:`NoiseModel`.
 * Fix conversion of symbols into qiskit.
 * Require qiskit >= 1.2.0.
 * Add conversion of controlled unitary gates from qiskit to tket.
--- a/pytket/extensions/qiskit/backends/aer.py
+++ b/pytket/extensions/qiskit/backends/aer.py
@@ -694,7 +694,7 @@ class AerDensityMatrixBackend(_AerBaseBackend):
     ) -> None:
         super().__init__()
         self._qiskit_backend = qiskit_aer_backend(self._qiskit_backend_name)
-        self._noise_model = noise_model
+
         gate_set = _tket_gate_set_from_qiskit_backend(self._qiskit_backend).union(
             self._allowed_special_gates
         )
--- a/pytket/extensions/qiskit/backends/aer.py
+++ b/pytket/extensions/qiskit/backends/aer.py
@@ -304,6 +304,8 @@ class _AerBaseBackend(Backend):
                     qc.save_state()
                 elif self.supports_unitary:
                     qc.save_unitary()
+                elif self.supports_density_matrix:
+                    qc.save_state()
                 qcs.append(qc)
                 tkc_qubits_count.append(c0.n_qubits)
                 ppcirc_strs.append(json.dumps(ppcirc_rep))
@@ -662,6 +664,7 @@ class AerUnitaryBackend(_AerBaseBackend):
 
 class AerDensityMatrixBackend(_AerBaseBackend):
     _supports_density_matrix = True
+    _supports_state = False
     _memory = False
     _noise_model = None
     _needs_transpile = True
@@ -669,21 +672,22 @@ class AerDensityMatrixBackend(_AerBaseBackend):
 
     _qiskit_backend_name = "aer_simulator_density_matrix"
 
-    def __init__(self, n_qubits: int = 40) -> None:
+    def __init__(
+        self, n_qubits: int = 40, noise_model: Optional[NoiseModel] = None
+    ) -> None:
         super().__init__()
         self._qiskit_backend = qiskit_aer_backend(self._qiskit_backend_name)
+        self._noise_model = noise_model
         self._backend_info = BackendInfo(
             name=type(self).__name__,
             device_name=self._qiskit_backend_name,
             version=__extension_version__,
             architecture=FullyConnected(n_qubits),
             gate_set=_tket_gate_set_from_qiskit_backend(self._qiskit_backend),
-            supports_midcircuit_measurement=True,  # is this correct?
+            supports_midcircuit_measurement=True,
             misc={"characterisation": None},
         )
         self._required_predicates = [
-            NoClassicalControlPredicate(),
-            NoFastFeedforwardPredicate(),
             GateSetPredicate(self._backend_info.gate_set),
         ]
 
--- a/pytket/extensions/qiskit/backends/aer.py
+++ b/pytket/extensions/qiskit/backends/aer.py
@@ -680,6 +680,13 @@ class AerDensityMatrixBackend(_AerBaseBackend):
 
     _qiskit_backend_name = "aer_simulator_density_matrix"
 
+    _allowed_special_gates = {
+        OpType.Measure,
+        OpType.Barrier,
+        OpType.Reset,
+        OpType.RangePredicate,
+    }
+
     def __init__(
         self,
         noise_model: Optional[NoiseModel] = None,
@@ -688,6 +695,17 @@ class AerDensityMatrixBackend(_AerBaseBackend):
         super().__init__()
         self._qiskit_backend = qiskit_aer_backend(self._qiskit_backend_name)
         self._noise_model = noise_model
+        gate_set = _tket_gate_set_from_qiskit_backend(self._qiskit_backend).union(
+            self._allowed_special_gates
+        )
+        self._noise_model = _map_trivial_noise_model_to_none(noise_model)
+        characterisation = _get_characterisation_of_noise_model(
+            self._noise_model, gate_set
+        )
+        self._has_arch = bool(characterisation.architecture) and bool(
+            characterisation.architecture.nodes
+        )
+
         self._backend_info = BackendInfo(
             name=type(self).__name__,
             device_name=self._qiskit_backend_name,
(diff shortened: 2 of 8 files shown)


docs/changelog.rst
```diff
@@ -1,9 +1,13 @@
 Changelog
 ~~~~~~~~~
 
+.. currentmodule:: pytket.extensions.qiskit
```
sjdilkes, line 4: Is this deliberate? I'm unfamiliar with the syntax

pytket/extensions/qiskit/backends/aer.py
```diff
@@ -660,6 +663,71 @@ def __init__(self, n_qubits: int = 40) -> None:
         ]
 
 
+class AerDensityMatrixBackend(_AerBaseBackend):
+    """
+    Backend for running simulations on the Qiskit Aer density matrix simulator.
+
+    :param noise_model: Noise model to apply during simulation. Defaults to None.
```
sjdilkes, line 670: Should these be defined here or in the init? I'm not sure what our standard is

pytket/extensions/qiskit/backends/aer.py
```diff
@@ -660,6 +663,71 @@ def __init__(self, n_qubits: int = 40) -> None:
...
+    :param n_qubits: The maximum number of qubits supported by the backend.
+    """
+
+    _supports_density_matrix = True
+    _supports_state = False
+    _memory = False
+    _noise_model = None
```
sjdilkes, line 677: type information or all these attributes would be helpful

pytket/extensions/qiskit/backends/aer.py
```diff
@@ -660,6 +663,71 @@ def __init__(self, n_qubits: int = 40) -> None:
...
+
+    def __init__(
+        self,
+        noise_model: Optional[NoiseModel] = None,
+        n_qubits: int = 40,
+    ) -> None:
+        super().__init__()
+        self._qiskit_backend = qiskit_aer_backend(self._qiskit_backend_name)
+        self._noise_model = noise_model
+        gate_set = _tket_gate_set_from_qiskit_backend(self._qiskit_backend).union(
```
sjdilkes, line 698: type for this would be great

pytket/extensions/qiskit/backends/aer.py
```diff
@@ -660,6 +663,71 @@ def __init__(self, n_qubits: int = 40) -> None:
...
+        )
+        self._has_arch = bool(characterisation.architecture) and bool(
+            characterisation.architecture.nodes
+        )
+
+        self._backend_info = BackendInfo(
+            name=type(self).__name__,
+            device_name=self._qiskit_backend_name,
+            version=__extension_version__,
+            architecture=FullyConnected(n_qubits),
```
sjdilkes, line 713: Should this switch between `characterisation.architecture` and `FullyConnected`? Though maybe using an architecture with a density matrix is not likely behaviour anyway?

pytket/extensions/qiskit/backends/aer.py
```diff
@@ -660,6 +663,71 @@ def __init__(self, n_qubits: int = 40) -> None:
...
+            gate_set=_tket_gate_set_from_qiskit_backend(self._qiskit_backend),
+            supports_midcircuit_measurement=True,
+            supports_reset=True,
+            supports_fast_feedforward=True,
```
sjdilkes, line 717: This is definitely true for the density matrix simulator?

pytket/extensions/qiskit/result_convert.py
```diff
@@ -133,13 +133,17 @@ def qiskit_experimentresult_to_backendresult(
...
+            density_matrix = datadict["density_matrix"].reverse_qargs().data
+
     return BackendResult(
         c_bits=c_bits,
         q_bits=q_bits,
         shots=shots,
         counts=counts,
         state=state,
         unitary=unitary,
+        density_matrix=density_matrix,
```
sjdilkes, line 146: `BackendResult` already has a density_matrix attribute?

sjdilkes (changes requested):

docs/changelog.rst
CalMacCQ: Yeah, basically it means that the refernces to classes and functions don't need to be fully qualified. The link is also generated

i.e.

```
:py:class:`IBMQBackend`
```

vs 

```
:py:class:`~pytket.extensions.qiskit.IBMQBackend`
```

pytket/extensions/qiskit/backends/aer.py
CalMacCQ: added in https://github.com/CQCL/pytket-qiskit/pull/380/commits/1d5fd8c8871e7393745416e91c933d2d29bbd7ad

pytket/extensions/qiskit/backends/aer.py
CalMacCQ: added in https://github.com/CQCL/pytket-qiskit/pull/380/commits/1d5fd8c8871e7393745416e91c933d2d29bbd7ad

pytket/extensions/qiskit/backends/aer.py
CalMacCQ: Seems to me that we document this in the class docstring rather than the `__init__` method. I don't have a strong prefernce either way.

pytket/extensions/qiskit/backends/aer.py
CalMacCQ: I believe so. I have a test case for this [here](https://github.com/CQCL/pytket-qiskit/pull/380/files#diff-90b322f046055083e0af3af8ce16264e3f806fcac3c4e6b31b6a6687d891a629R1453)

pytket/extensions/qiskit/result_convert.py
CalMacCQ: I'm just passing it in as an arg when defining the `BackendResult` though right?

tests/backend_test.py
```diff
@@ -1430,3 +1432,63 @@ def test_ibmq_local_emulator(
...
+    output_state = np.array([0] * 7 + [1])
+
+    result1 = density_matrix_backend.run_circuit(circ1)
+
+    assert result1.get_density_matrix().shape == (8, 8)
+    assert np.allclose(
+        result1.get_density_matrix(), np.outer(output_state, output_state.conj())
+    )
+    # Example with resets and conditional gates
+    # Deterministic if input is a computational basis state
```
cqc-alec, line 1455: Don't quite understand this comment. What exactly is deterministic?

tests/backend_test.py
```diff
@@ -1430,3 +1432,63 @@ def test_ibmq_local_emulator(
...
+    circ = Circuit(3)
+    circ.X(0)
+    circ.X(1)
+    circ.SX(1)
+    circ.CZ(0, 1)
+    circ.CZ(1, 2)
+    assert noisy_density_sim.valid_circuit(circ)
+
+    result = noisy_density_sim.run_circuit(circ)
+    assert result.get_density_matrix().shape == (8, 8)
```
cqc-alec, line 1494: Can we also check that the density matrix differs from what we get with a non-noisy simulation?

pytket/extensions/qiskit/backends/aer.py
```diff
@@ -300,10 +300,13 @@ def process_circuits(
                     c0, ppcirc_rep = tkc, None
 
                 qc = tk_to_qiskit(c0, replace_implicit_swaps)
-                if self.supports_state:
+
+                if self.supports_state or self.supports_density_matrix:
                     qc.save_state()
```
cqc-alec, line 305: Shouldn't we be using https://qiskit.github.io/qiskit-aer/stubs/qiskit_aer.library.save_density_matrix.html ?

CalMacCQ: Yes, the `QuantumCircuit.save_state` gave equivalent answers but I think you're right. Its more correct to save the density matrix.

Thanks

tests/backend_test.py
CalMacCQ: The way I've done this so far is to check the purity of the state represented by the density matrix

$$
\gamma = \text{Tr}(\rho^2)
$$

If $\gamma =1$, the state is pure. If $\gamma < 1$ the state is mixed. I've verified in [6d00b85](https://github.com/CQCL/pytket-qiskit/pull/380/commits/6d00b85ca6cf7e90cc5cec04aa21bcac566672c5) that the density matrix given by the noisy simulator is a mixed state and the noisless a pure state.

I'm not sure whether this is thorough enough. Perhaps its a good idea to calcluate the expectation value of some observable $E$ and verify that they are not equal.

tests/backend_test.py
CalMacCQ: So despite the fact that the circuit contains measures and resets it prepares a state deterministically assuming that it starts from the $|0\rangle^{\otimes n}$ state.

tests/backend_test.py
```diff
@@ -1430,3 +1432,69 @@ def test_ibmq_local_emulator(
...
+    circ.SX(1)
+    circ.CZ(0, 1)
+    circ.CZ(1, 2)
+    assert noisy_density_sim.valid_circuit(circ)
+
+    result = noisy_density_sim.run_circuit(circ)
+    noisy_dm = result.get_density_matrix()
+    assert noisy_dm.shape == (8, 8)
+    # Check purity to verify mixed state
+    assert np.trace(noisy_dm**2).real < 1
```
cqc-alec, line 1500: Could we change this to "< 0.99" or something so that we rule out the possibility of rounding errors causing the test to pass when it shouldn't?

tests/backend_test.py
```diff
@@ -1430,3 +1432,69 @@ def test_ibmq_local_emulator(
...
+    circ2.Reset(2)
+
+    result2 = density_matrix_backend.run_circuit(circ2)
+    assert result1.get_density_matrix().shape == (8, 8)
+    state_backend = AerStateBackend()
+    statevector = state_backend.run_circuit(circ2).get_state()
+    noiseless_dm2 = result2.get_density_matrix()
+    assert np.allclose(noiseless_dm2, np.outer(statevector, statevector.conj()))
+    # Check purity to verify that we have a pure state
+    assert np.trace(noiseless_dm2**2).real == 1
```
cqc-alec, line 1473: Better to use `np.isclose()` for floating-point comparison, to account for rounding error.

cqc-alec (commented):
Couple of small comments.

tests/backend_test.py
CalMacCQ: Good point, done in [ff91d0a](https://github.com/CQCL/pytket-qiskit/pull/380/commits/ff91d0a4c9ad9116b4e63bfdd0d498591808eeec)

tests/backend_test.py
CalMacCQ: Thanks, done.

sjdilkes (approved):

PR #383, opened 2024-08-27
chore: use python 3.10+ types, cleanup docs

### LINKED ISSUE #347: Update Sphinx linking with intersphinx and autodoc typehints
* Use the intersphinx extension properly to link from the pytket-qiskit API docs to other places.
* The `:type <param>:` and `:rtype:` in source are redundant if we use sphinx autodoc type hints where types are inferred from the source. Types don't need to be written down twice if we use this extension.

separate note: we still use a lot of `List` vs `list` and `Tuple` vs `tuple` for type annotations. Now that we only support python >3.10 we can update these and not have to import from the typing module.

CalMacCQ:
# Description

Used python 3.10+ types everywhere now that only 3.10+ versions of python are supported.

Some examples
* use `|` rather than `Union`
* using `list` rather than `typing.List`. `dict` rather than `typing.Dict` etc.

If we want to do this for other repos we can use [pyupgrade](https://github.com/asottile/pyupgrade) to automate most of this.

Also removed some redundant type declarations for sphinx e.g. `:type` and `:rtype`. the [sphinx-autodoc-typehints](https://github.com/tox-dev/sphinx-autodoc-typehints) extension infers the types from the function signature so we can remove a little duplication.

Not very important to do this but it was quick and easy to do.

Also cleaned up [this docs page](https://tket.quantinuum.com/extensions/pytket-qiskit/) a bit. Making more use of sphinx crossreferencing.

# Related issues

closes #347

# Checklist

### DIFF
--- a/.github/workflows/docs/conf.py
+++ b/.github/workflows/docs/conf.py
@@ -113,7 +113,7 @@ def correct_signature(
     what: str,
     name: str,
     obj: Any,
-    options: 
+    options: dict,
     signature: str,
     return_annotation: str,
 ) -> (str, str):
--- a/.github/workflows/docs/conf.py
+++ b/.github/workflows/docs/conf.py
@@ -48,7 +48,7 @@ autodoc_member_order = "groupwise"
 # The following code is for resolving broken hyperlinks in the doc.
 
 import re
-from typing import Any, Dict, List, Optional
+from typing import Any, Optional
 from urllib.parse import urljoin
 
 from docutils import nodes
--- a/.github/workflows/docs/conf.py
+++ b/.github/workflows/docs/conf.py
@@ -113,7 +113,7 @@ def correct_signature(
     what: str,
     name: str,
     obj: Any,
-    options: Dict,
+    options: 
     signature: str,
     return_annotation: str,
 ) -> (str, str):
--- a/docs/intro.txt
+++ b/docs/intro.txt
@@ -95,7 +95,7 @@ to disk:
 
     QiskitRuntimeService.save_account(channel="ibm_quantum", token=ibm_token, instance=inst)
 
-To see which devices you can access you can use the :py:meth:`~pytket.extensions.qiskit.IBMQBackend.available_devices` method on the :py:class:`IBMQBackend`. Note that it is possible to pass an optional ``instance`` argument to this method. This allows you to see which IBM devices are accessible with your credentials.
+To see which devices you can access, use the :py:meth:`IBMQBackend.available_devices` method. Note that it is possible to pass an optional ``instance`` argument to this method. This allows you to see which IBM devices are accessible with your credentials.
 
 ::
 
@@ -106,7 +106,7 @@ To see which devices you can access you can use the :py:meth:`~pytket.extensions
     backendinfo_list = backend.available_devices(instance=my_instance) 
     print([backend.device_name for backend in backendinfo_list])
 
-For more information see the documentation for `qiskit-ibm-runtime <https://docs.quantum.ibm.com/api/qiskit-ibm-runtime>`_.
+For more information, see the documentation for `qiskit-ibm-runtime <https://docs.quantum.ibm.com/api/qiskit-ibm-runtime>`_.
 
 
 Method 2: Saving credentials in a local pytket config file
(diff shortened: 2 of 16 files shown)


CalMacCQ:
Couple mistakes here. Will finish it off.

.github/workflows/docs/conf.py
```diff
@@ -48,7 +48,7 @@
 # The following code is for resolving broken hyperlinks in the doc.
 
 import re
-from typing import Any, Dict, List, Optional
+from typing import Any, Optional
 from urllib.parse import urljoin
```
CalMacCQ, line 52: This file will likely be removed soon with the next batch of docs updates

pytket/extensions/qiskit/backends/aer.py
```diff
@@ -236,7 +226,7 @@ def _arch_independent_default_compilation_pass(
         return SequencePass([DecomposeBoxes(), FullPeepholeOptimise()])
 
     def default_compilation_pass(
-        self, optimisation_level: int = 2, placement_options: Optional[Dict] = None
+        self, optimisation_level: int = 2, placement_options: Optional[dict] = None
```
CalMacCQ, line 229: These `Optional[dict]` annotations are actually incomplete. We should update the types of the keys and values in a separate PR.

CalMacCQ: fixed the incomplete dict typing in [8ee1d85](https://github.com/CQCL/pytket-qiskit/pull/383/commits/8ee1d852d41e743c532d9df0cf44fbef92b1f900) and [d93d126](https://github.com/CQCL/pytket-qiskit/pull/383/commits/d93d126a9876baae22b1c3e448ebf37340722737)

docs/intro.txt
```diff
@@ -120,7 +121,7 @@ Alternatively, you can store your credentials in local pytket config using the :
 
 After saving your credentials you can access ``pytket-qiskit`` backend repeatedly without having to re-initialise your credentials.
 
-If you are a member of an IBM hub then you can add this information to ``set_ibmq_config`` as well.
+If you are a member of an IBM hub then you can add this information to :py:meth:`~pytket.extensions.qiskit.backends.config.set_ibmq_config` as well.
```
CalMacCQ, line 124: Crossreferncing this way means the link to the function/class is generated in the html page.

docs/intro.txt
```diff
@@ -75,12 +75,13 @@ In this section we are assuming that you have set the following variables with t
...
 
     # Replace the placeholders with your actual values
+
     ibm_token = '<your_ibm_token_here>'    
     hub = '<your_hub_here>'
     group = '<your_group_here>'
     project = '<your_project_here>'
 
-    my_instance=f"{hub}/{group}/{project}"
+    inst = f"{hub}/{group}/{project}"
```
CalMacCQ, line 84: This rename is just so that the `QiskitRuntimeService` cell doesn't need a scrollbar and shows all of the contents.

pytket/extensions/qiskit/backends/ibmq_emulator.py
```diff
@@ -14,26 +14,21 @@
...
-)
+from pytket.backends.backend import Backend
+from pytket.backends.status import CircuitStatus
+
 from pytket.backends.backendinfo import BackendInfo
 from pytket.backends.backendresult import BackendResult
-from pytket.backends.resulthandle import _ResultIdTuple
-from pytket.circuit import Circuit
+from pytket.backends.resulthandle import _ResultIdTuple, ResultHandle
+from pytket._tket.circuit import Circuit
```
cqc-alec, line 31: Not sure why this change is needed?

cqc-alec (commented):
Thanks, looks good, one small query.

pytket/extensions/qiskit/backends/ibmq_emulator.py
CalMacCQ: This just stops the typechecker in my IDE from complaining. Can change it back if preferred.

cqc-alec: OK, I think it's better to revert it; we shouldn't need to use `_tket` in the extensions code I think.

CalMacCQ: No problem, done in [ca9e1a8](https://github.com/CQCL/pytket-qiskit/pull/383/commits/ca9e1a86f0b6600f1e2b8f3ca35a8d0c663f23bc)

cqc-alec (approved):

Issue #350, opened 2024-06-14
DeprecationWarning: V1 fake backends 


cqc-melf:
DeprecationWarning: V1 fake backends are deprecated as of qiskit-ibm-runtime 0.24 and will be removed no sooner than 3 months after the release date. Please use V2 fake backends instead.

PR #385, opened 2024-09-06
Remove some deprecation warnings

cqc-alec:
Another step towards #350 ...

I think the only deprecation warnings now are from `TketBackend`, which is not so easy to update and may require a complete rewrite ...

I deleted a couple of tests that use `qiskit_algorithms` which depended on the deprecated classes. There is a similar example in the user manual. I've checked that it still works.

### DIFF
--- a/pytket/extensions/qiskit/backends/ibm.py
+++ b/pytket/extensions/qiskit/backends/ibm.py
@@ -43,7 +43,8 @@ from qiskit_ibm_runtime import (  # type: ignore
     SamplerV2,
     RuntimeJob,
 )
-from qiskit.providers.models import BackendProperties, QasmBackendConfiguration  # type: ignore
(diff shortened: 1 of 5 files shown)


pytket/extensions/qiskit/backends/ibm.py
```diff
@@ -43,7 +43,8 @@
     SamplerV2,
     RuntimeJob,
 )
-from qiskit.providers.models import BackendProperties, QasmBackendConfiguration  # type: ignore
+from qiskit_ibm_runtime.models.backend_configuration import PulseBackendConfiguration  # type: ignore
```
yao-cqc, line 46: Why replacing `QasmBackendConfiguration` with `PulseBackendConfiguration`?

cqc-alec: Since we switched to using qiskit runtime, this is actually the type of what `IBMBackend.configuration()` returns. This change only corrects the typing annotations.

PR #389, opened 2024-09-09
Update `TketBackend` to derive from `BackendV2`

### LINKED ISSUE #350: DeprecationWarning: V1 fake backends

cqc-alec:
Closes #350 .

This change would break [one of the examples](https://docs.quantinuum.com/tket/user-guide/manual/manual_backend.html#embedding-into-qiskit) in the manual, and perhaps also one or two places in the example notebooks (I haven't checked). This shouldn't affect the docs build since it's pinned to a specific version of pytket-qiskit, but we should probably revise those sections. It seems that the relevant parts of `qiskit_algorithms` (a community-maintained project) haven't yet been updated to use the new backend class.

This PR removes the last of the deprecation warnings from the qiskit tests that were coming from our code.

### DIFF
--- a/pytket/extensions/qiskit/tket_backend.py
+++ b/pytket/extensions/qiskit/tket_backend.py
@@ -31,8 +31,10 @@ from pytket.architecture import FullyConnected
 def _extract_basis_gates(backend: Backend) -> list[str]:
     standard_gate_mapping = get_standard_gate_name_mapping()
     for pred in backend.required_predicates:
-        if type(pred) == GateSetPredicate:
+        if type(pred) is GateSetPredicate:
             basis_gates = []
+            # Restrict to the gate set accepted by the backend, the converters,
+            # and the Target.from_configuration() method.
             for optype in pred.gate_set:
                 if optype in _gate_str_2_optype_rev.keys():
                     gate_name = _gate_str_2_optype_rev[optype]
--- a/pytket/extensions/qiskit/tket_backend.py
+++ b/pytket/extensions/qiskit/tket_backend.py
@@ -12,36 +12,39 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import inspect
 from typing import Optional, Any
+from qiskit.circuit.library.standard_gates import get_standard_gate_name_mapping  # type: ignore
 from qiskit.circuit.quantumcircuit import QuantumCircuit  # type: ignore
-from qiskit.providers.backend import BackendV1  # type: ignore
-from qiskit.providers.models import QasmBackendConfiguration  # type: ignore
+from qiskit.providers.backend import BackendV2  # type: ignore
 from qiskit.providers import Options  # type: ignore
+from qiskit.transpiler import CouplingMap, Target  # type: ignore
 from pytket.extensions.qiskit import AerStateBackend, AerUnitaryBackend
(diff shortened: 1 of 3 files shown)


pytket/extensions/qiskit/tket_backend.py
```diff
@@ -12,36 +12,39 @@
...
-                for optype in pred.gate_set
-                if optype in _gate_str_2_optype_rev.keys()
-            ]
+            basis_gates = []
+            for optype in pred.gate_set:
+                if optype in _gate_str_2_optype_rev.keys():
+                    gate_name = _gate_str_2_optype_rev[optype]
+                    if gate_name in standard_gate_mapping:
+                        gate_obj = standard_gate_mapping[gate_name]
+                        if gate_obj.num_qubits in [1, 2] or inspect.isclass(gate_obj):
```
cqc-alec, line 41: See https://github.com/Qiskit/qiskit/blob/2ef371ae0d159a6dfd643805f3e5e5fdec37ab88/qiskit/transpiler/target.py#L1066 -- have to restrict the gateset to match what is accepted by the `Target` method.

pytket/extensions/qiskit/tket_backend.py
```diff
@@ -12,36 +12,39 @@
...
     for pred in backend.required_predicates:
         if type(pred) == GateSetPredicate:
-            return [
-                _gate_str_2_optype_rev[optype]
-                for optype in pred.gate_set
-                if optype in _gate_str_2_optype_rev.keys()
-            ]
+            basis_gates = []
+            for optype in pred.gate_set:
+                if optype in _gate_str_2_optype_rev.keys():
```
CalMacCQ, line 37: Perhaps worth adding a comment explaining what these "if" statements are for?

pytket/extensions/qiskit/tket_backend.py
```diff
@@ -12,36 +12,39 @@
...
-    CompilationUnit,
-)
+from pytket.predicates import GateSetPredicate, CompilationUnit
 from pytket.architecture import FullyConnected
 
 
 def _extract_basis_gates(backend: Backend) -> list[str]:
+    standard_gate_mapping = get_standard_gate_name_mapping()
     for pred in backend.required_predicates:
         if type(pred) == GateSetPredicate:
```
CalMacCQ, line 34: Rather pedantic but is there a reason not to use "is" or "isinstance" here vs "=="?

CalMacCQ (commented):
Looks good :) just a few very minor comments

pytket/extensions/qiskit/tket_backend.py
CalMacCQ: Do I understand correctly that the `gate_obj` variable is not always a class? Presumably thats why the inspect check is needed.

cqc-alec: Yes, some gates such as CCX were rejected when I omitted this check. I am not sure when `gate_obj` is a class, but simply copied that from the code linked above.

CalMacCQ (approved):
Thanks :)

PR #382, opened 2024-08-24
refactor: Simplify some internals of `qiskit_to_tk` conversion

CalMacCQ:
# Description

This PR begins a refactor of the `qiskit_to_tk` and `tk_to_qiskit` converters. Here I've focussed on refactoring some of the large `CircuitBuilder.add_qiskit_data` method (used as part of `qiskit_to_tk` into separate internal functions. I've added some short docstrings so its hopefully clearer whats going on.

There's still a lot to improve in the conversion code but I think it makes sense to do this refactor in several stages to make small fixes like https://github.com/CQCL/pytket-qiskit/issues/366 a bit easier.

I've avoided imporvements which involve user facing changes for now. Will add these in a follow up.

# Related issues

#313 

# Checklist

- [x] I have performed a self-review of my code.
- [x] I have commented hard-to-understand parts of my code.
- [ ] I have made corresponding changes to the public API documentation.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] I have updated the changelog with any user-facing changes.

PR #372, opened 2024-08-13
Add conversion of controlled unitary gates from Qiskit to TKET.

gkpotter:
# Description

Add support for converting controlled unitary gates from Qiskit to TKET.

# Related issues

Please mention any github issues addressed by this PR.

# Checklist

- [x] I have performed a self-review of my code.
- [x] I have commented hard-to-understand parts of my code.
- [ ] I have made corresponding changes to the public API documentation.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] I have updated the changelog with any user-facing changes.

# Notes

I have yet to update the changelog/API docs, in case more changes or refactoring are required.

Issue #313, opened 2024-04-05
Idea: Refactor `qiskit_to_tk` and `tk_to_qiskit`


CalMacCQ:
The [qiskit_to_tk](https://tket.quantinuum.com/extensions/pytket-qiskit/api.html#pytket.extensions.qiskit.qiskit_to_tk) and [tk_to_qiskit](https://tket.quantinuum.com/extensions/pytket-qiskit/api.html#pytket.extensions.qiskit.tk_to_qiskit) converters are a frequent source of bug reports and feature requests from users. I suspect for a significant number of people these converters are their first exposure to using pytket and related libraries.

The code for doing the conversion is old and has been patched together over time. Its now about [1000 lines of untidy python](https://github.com/CQCL/pytket-qiskit/blob/develop/pytket/extensions/qiskit/qiskit_convert.py). I think it could make the code simpler and more maintainable if we rethink/refactor the converters perhaps splitting out some of the compilcated case handling into separate internal functions.

I think this would save us time in the long run when addressing bugs and feature requests from users. May also make it easier for new contributors to make PRs.

A couple of thoughts...
* pytket and extensions now support python 3.10, 3.11 and 3.12. This could allow us to make use of [structural pattern matching](https://peps.python.org/pep-0636/) (introduced in 3.10) in the circuit parsers.
* I think having circuit converters which handle higher level circuit boxes is valuable as it allows easier (and fairer) side by side benchmarking of pytket vs qiskit. Also it allows qiskit users to try out pytket with less hassle. One could of course trivially just convert a pytket circuit to a qiskit circuit by exporting to and from QASM but this translation would be very lossy and remove abstractions which can be useful for the compiler. Handling more ops also expands the range of circuits we can run on the idealised `AerBackend` simulator.

CalMacCQ:
Closed this for now as I don't have any plans for imporving this code more in the immediate future beyond the changes in #382 and #391. However theres still a lot of untidy conversion code especially in the `tk_to_qiskit` direction.

Maybe worth coming back to this at some point.

PR #391, opened 2024-09-12
feat: Improved handling of controlled gates in converters

### LINKED ISSUE #178: Use improved `QControlBox` to handle control states in `qiskit_to_tk`
As of pytket 1.19.0 we now support custom control states in `QControlBox`.

Currently if there is a qiskit `ControlledGate` controled on some bitstring that isn't "all |1>" we insert X's to handle this. See https://github.com/CQCL/pytket-qiskit/issues/117.

We should now be able to handle the qiskit control states directly without having to insert Pauli X gates on either side.

CalMacCQ:
# Description

This PR adds support for handling the `control_state` of a controlled operation in the `qiskit_to_tk` and `tk_to_qiskit` converters. The control state is now handled directly instead of adding `X` gates on the control qubits.

I've also taken the liberty of refactoring out the "`QControlBox` building" from the `CircuitBuildier.add_qiskit_data` method in the same spirit as  #382 

I've also reworked some internals of how unitary gates are converted to and fro. Handling of controlled unitary boxes was added in #372 . ~I think it'd be fairly easy to add the other direction into this PR too (see #378)~ (I think its best to do this separately).

Original PR adding control state handling with X gates -> https://github.com/CQCL/pytket-qiskit/pull/118

# Related issues
#378 
#313 
closes #178 

# Checklist

### DIFF
--- a/pytket/extensions/qiskit/qiskit_convert.py
+++ b/pytket/extensions/qiskit/qiskit_convert.py
@@ -82,7 +82,11 @@ from pytket.circuit import (
 from pytket.unit_id import _TEMP_BIT_NAME
 from pytket.pauli import Pauli, QubitPauliString
 from pytket.architecture import Architecture, FullyConnected
-from pytket.utils import QubitPauliOperator, gen_term_sequence_circuit
+from pytket.utils import (
+    QubitPauliOperator,
+    gen_term_sequence_circuit,
+    permute_rows_cols_in_unitary,
+)
 from pytket.passes import AutoRebase
 
 if TYPE_CHECKING:
@@ -361,16 +365,19 @@ def _get_unitary_box(u_gate: UnitaryGate) -> UnitaryBox:
     assert len(params) == 1
     unitary = cast(NDArray[np.complex128], params[0])
     assert isinstance(u_gate.num_qubits, int)
+    pytket_unitary: NDArray[np.complex128] = permute_rows_cols_in_unitary(
+        unitary, tuple(reversed(range(u_gate.num_qubits)))
+    )
     match u_gate.num_qubits:
         case 1:
             assert unitary.shape == (2, 2)
-            return Unitary1qBox(unitary)
+            return Unitary1qBox(pytket_unitary)
         case 2:
             assert unitary.shape == (4, 4)
-            return Unitary2qBox(unitary)
+            return Unitary2qBox(pytket_unitary)
         case 3:
             assert unitary.shape == (8, 8)
-            return Unitary3qBox(unitary)
+            return Unitary3qBox(pytket_unitary)
         case _:
             raise NotImplementedError(
                 f"Conversion of {u_gate.num_qubits}-qubit unitary gates not supported."
--- a/pytket/extensions/qiskit/qiskit_convert.py
+++ b/pytket/extensions/qiskit/qiskit_convert.py
@@ -538,8 +538,6 @@ class CircuitBuilder:
 
             elif type(instr) is UnitaryGate:
                 unitary = cast(NDArray[np.complex128], instr.params[0])
-                dim: int = 2**instr.num_qubits
-                assert unitary.shape == (dim, dim)
                 if len(qubits) == 0:
                     # If the UnitaryGate acts on no qubits, we add a phase.
                     self.tkc.add_phase(np.angle(unitary[0][0]) / np.pi)
--- a/pytket/extensions/qiskit/qiskit_convert.py
+++ b/pytket/extensions/qiskit/qiskit_convert.py
@@ -614,6 +614,10 @@ def qiskit_to_tk(qcirc: QuantumCircuit, preserve_param_uuid: bool = False) -> Ci
     return builder.circuit()
 
 
+def _get_qiskit_control_state(bool_list: list[bool]) -> str:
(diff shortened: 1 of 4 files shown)


tests/qiskit_convert_test.py
```diff
@@ -865,6 +867,21 @@ def test_controlled_unitary_conversion() -> None:
     assert np.allclose(u_qc, u_tkc)
 
 
+def test_qcontrol_box_conversion_to_qiskit() -> None:
```
CalMacCQ, line 870: I think I will add tests for a wider range of controlled operations

pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -290,9 +294,27 @@ def _string_to_circuit(
     return circ
 
 
+def _get_pytket_ctrl_state(bitstring: str, n_bits: int) -> tuple[bool, ...]:
+    "Converts a little endian string '001'=1 (LE) to (1, 0, 0)."
```
cqc-alec, line 298: That looks like a big-endian string if the value is 1. But you are just reversing it. Could implement this as:
```python
    return tuple(reversed(list(map(int, bitstr))))
```
Don't need the `n_bits` parameter.

tests/qiskit_convert_test.py
cqc-alec: Yes please.

pytket/extensions/qiskit/qiskit_convert.py
CalMacCQ: The problem is that we don't know the number of control qubits in advance. 

We could have a binary encoding of "7" in 3 bits or 4 bits or a larger number. We still need to produce a list of the correct length to pass to `QControlBox`

cqc-alec: OK, in that case can implement as
```python
return tuple(reversed(list(map(int, bitstr.zfill(n_bits)))))
```
?

CalMacCQ: The following works well for me (all tests pass) .

```python
def _get_pytket_ctrl_state(bitstring: str, n_bits: int) -> tuple[bool, ...]:
    assert set(bitstring).issubset({"0", "1"})
    new_bitstring = bitstring.zfill(n_bits)
    pytket_ctrl_state = reversed([bool(int(b)) for b in new_bitstring])
    return tuple(pytket_ctrl_state)
```

The snippet above didn't work as a drop in but I think its a case of the zeros being added to the right rather than the left.

I didn't know about `str.z_fill()` thanks.

tests/qiskit_convert_test.py
```diff
@@ -865,6 +867,32 @@ def test_controlled_unitary_conversion() -> None:
...
+    circ1.add_gate(ccch_001, [0, 1, 2, 3])
+    circ1.add_gate(cccs_110, [0, 1, 2, 3])
+    circ1.add_gate(cccRy_100, [3, 2, 1, 0])
+    circ1.add_gate(ccU3_10, [1, 0, 2])
+    qc = tk_to_qiskit(circ1)
+    circ2 = qiskit_to_tk(qc)
+    DecomposeBoxes().apply(circ1)
+    DecomposeBoxes().apply(circ2)
+    assert circ1 == circ2
+    assert compare_unitaries(circ1.get_unitary(), circ2.get_unitary())
```
CalMacCQ, line 893: I think this check of unitary equivalence may be redundant given the previous line

tests/qiskit_convert_test.py
CalMacCQ: I've fixed the conversion of parametrised gates in https://github.com/CQCL/pytket-qiskit/pull/391/commits/b42946e54230c61bb9b65037f5ae6e31f7b19ab2 and added some controlled parametrised gates in https://github.com/CQCL/pytket-qiskit/pull/391/commits/c1b315cbacf7e66de9cf34cbf7c41358c2652a98

tests/qiskit_convert_test.py
```diff
@@ -865,6 +867,21 @@ def test_controlled_unitary_conversion() -> None:
     assert np.allclose(u_qc, u_tkc)
 
 
+def test_qcontrol_box_conversion_to_qiskit() -> None:
+    h_op = Op.create(OpType.H)
+    multi_controlled_h = QControlBox(
+        h_op, n_controls=3, control_state=(False, False, True)
+    )
+    circ1 = Circuit(4, name="CCH test")
```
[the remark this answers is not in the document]
CalMacCQ: Changed the name. There are now more gates in the test circuit.

tests/qiskit_convert_test.py
cqc-alec: Yes, can remove this. However, I think we should add a test that compares the unitary of the tket circuit with the unitary of the qiskit circuit (adjusted for endianness).

cqc-alec (changes requested):

tests/qiskit_convert_test.py
CalMacCQ: added in [05dd936](https://github.com/CQCL/pytket-qiskit/pull/391/commits/05dd9361fa879c7cf713094310423e5081c63077)

cqc-alec (approved):

PR #392, opened 2024-09-17
docs: add `pytket-docs-theming` submodule

CalMacCQ:
# Description

Essentially a clone of https://github.com/CQCL/pytket-cutensornet/pull/153 but for pytket-qiskit.

Using the [pytket-docs-theming](https://github.com/CQCL/pytket-docs-theming) repo as a source of truth for the different extension modules.

Also took the oppurtunity to clean up the legacy docs build.

Added scripts for installing the docs dependencies and building the docs (https://github.com/CQCL/pytket-qiskit/pull/392/commits/b0b55c9cad8cb2b41b1fd6bd598802a33f99d27f). These scripts will not be used in the website build... Only in this repository.

My local build below

Screenshot 2024-09-17 at 18 08 21

I note that theres a decent amount of duplication in the github workflows as the docs are built multiple times (see `docs.yml` and `build_and_test.yml`. Should we update these at this point? Or should that be done at a later time?

# Related issues

https://github.com/CQCL/pytket-docs/issues/360

# Checklist


.github/workflows/build_and_test.yml
```diff
@@ -80,6 +80,7 @@ jobs:
       if:  (matrix.os == 'ubuntu-22.04') && (github.event_name == 'pull_request' || github.event_name == 'schedule' )
       run: |
         cd docs && bash ./install.sh
+        poetry run pip install ../.
```
CalMacCQ, line 83: Not sure if this is how I should be installing the pytket-qiskit version in this workflow. 

Would it be better to install a `.whl` directly?

.github/workflows/build_and_test.yml
```diff
@@ -128,14 +132,16 @@ jobs:
         path: wheelhouse
     - name: Install pip, wheel
       run: pip install -U pip wheel
+    - name: Install poetry
+      run: pip install poetry
     - name: Install extension
       run: for w in `find wheelhouse/ -type f -name "*.whl"` ; do pip install $w ; done
```
CalMacCQ, line 138: I'm not sure what do do with this as I don't think its installed in the poetry env

.github/workflows/build_and_test.yml
cqc-melf: I think there should be a wheel available to install at this point.
Could you try something like below?
```
run: for w in `find wheelhouse/ -type f -name "*.whl"` ; do poetry install $w ; done
```
Or if that is not working
```
run: for w in `find wheelhouse/ -type f -name "*.whl"` ; do poetry run pip install $w ; done
```

.github/workflows/build_and_test.yml
cqc-melf: As above, I would replace this with:

cqc-melf, line 138:
```suggestion
      run: for w in `find wheelhouse/ -type f -name "*.whl"` ; do poetry install $w ; done

Comment (cqc-melf) [reply .github/workflows/build_and_test.yml:138]: (see above if this is not working)

File: .github/workflows/docs.yml [hunk 3]
```diff
@@ -14,24 +14,20 @@ jobs:
...
+      run: pip install poetry
     - name: Install docs dependencies
-      run: |
-        pip install -r .github/workflows/docs/requirements.txt
+      run: cd docs && bash install.sh 
+    - name: Install pytket qiskit
+      run: cd docs && poetry run pip install ../.
     - name: Test building docs
-      timeout-minutes: 20
+      timeout-minutes: 10
```
cqc-melf, line 30: Have you reduced the timeout on purpose?

cqc-melf (commented):
I have a few questions / suggestions.

I think it might be good to try a release candidate from this branch to check if this works?

.github/workflows/docs.yml
CalMacCQ: Yes, would you prefer I changed it back?

cqc-melf (approved):

PR #400, opened 2024-10-10
Add LightSABRE to default_compilation_pass

sjdilkes:
Replace `CXMappingPass` and `Placement` with a `CustomPass` that calls sabre routing as defined by `SabreLayoutPassManager`.

# Checklist


tests/backend_test.py
```diff
@@ -988,46 +988,6 @@ def lift_perm(p: dict[int, int]) -> np.ndarray:
     return pm
 
 
-@pytest.mark.skipif(skip_remote_tests, reason=REASON)
-def test_compilation_correctness(brisbane_backend: IBMQBackend) -> None:
```
sjdilkes, line 992: LightSabre Pass does not return any information regarding the implicit qubit permutation. If a circuit has `Measure` operations, as we'd expect for hardware, then this is not an issue, as all post-processing is completed based on output bits from `Measures`, which are still measuring the same states.

This test confirmed that the compilation, including SWAP operations, did not change the unitary, action, up to permutations, for a backend with some architecture.  In practice we do not expect users to want to do unitary simulation with qubit swapping, so I think we can reasonably remove this test.

CalMacCQ:
Small comment about docs.

Maybe we should update this section given that the default compilation pass is changing?
https://docs.quantinuum.com/tket/extensions/pytket-qiskit/#default-compilation

[Source file](https://github.com/CQCL/pytket-qiskit/blob/main/docs/index.md#default-compilation).

sjdilkes:
> Small comment about docs.
> 
> Maybe we should update this section given that the default compilation pass is changing? https://docs.quantinuum.com/tket/extensions/pytket-qiskit/#default-compilation
> 
> [Source file](https://github.com/CQCL/pytket-qiskit/blob/main/docs/index.md#default-compilation).

Good point - I've updated the source file. Please could you advice if there's anything else I need to do?

pytket/extensions/qiskit/backends/ibm_utils.py
```diff
@@ -70,3 +77,78 @@ def _batch_circuits(
         for n, indices in itertools.groupby(order, key=lambda i: n_shots[i])
     ]
     return batches, batch_order
+
+
+def _architecture_to_couplingmap(architecture: Architecture) -> CouplingMap:
+    """
+    Converts a pytket Architecture object to a Qiskit CouplingMap object.
+
+    :param architecture: Architecture to be converted
```
cqc-melf, line 86: Could you add some more details about the return value here?

pytket/extensions/qiskit/backends/ibm_utils.py
```diff
@@ -70,3 +77,78 @@ def _batch_circuits(
...
+    architecture: Architecture, optimization_level: int = 2, seed=0, attempts=20
+) -> Callable[[Circuit], Circuit]:
+    """
+    Generates a function that can be passed to CustomPass for running
+    LightSABRE routing.
+
+    :param architecture: Architecture LightSABRE routes circuits to match
+    :param optimization_level: Corresponds to qiskit optmization levels
+    :param seed: LightSABRE routing is stochastic, with this parameter setting the seed
+    :param attempts: Number of generated random solutions to pick from.
```
cqc-melf, line 114: Could you add some more details about the return value here?

tests/backend_test.py
cqc-melf: Is this potentially an issues for other cases? Should this be documented somewhere?

cqc-melf (commented):
Looks good, few questions.

pytket/extensions/qiskit/backends/ibm_utils.py
sjdilkes: done

tests/backend_test.py
sjdilkes: I've added this test back in, but edited it to only check the routing is correct

CalMacCQ:
> Good point - I've updated the source file. Please could you advice if there's anything else I need to do?


Local build of the docs looks good, no further edits needed.

tests/backend_test.py
```diff
@@ -1002,30 +1007,14 @@ def test_compilation_correctness(brisbane_backend: IBMQBackend) -> None:
     c.Rz(0.125, 2).X(2).Rz(0.25, 2)
     c.SX(3).Rz(0.125, 3).SX(3)
     c.CX(0, 3).CX(0, 4)
-    u_backend = AerUnitaryBackend()
     c.remove_blank_wires()
     FlattenRelabelRegistersPass().apply(c)
-    u = u_backend.run_circuit(c).get_unitary()
-    ibm_backend = brisbane_backend
+    c_pred = ConnectivityPredicate(brisbane_backend.backend_info.architecture)
```
cqc-melf, line 1012: This line still has an issues in the mypy check, could you please fix that or add an ignore?

cqc-melf: We should maybe add an issues to pytket that the predicate should accept fully connected, too?

cqc-melf:
```suggestion
    c_pred = ConnectivityPredicate(brisbane_backend.backend_info.architecture)  # type: ignore

sjdilkes: I've cast the object to `Architecture` which worked locally

cqc-melf:
The tests fail because of https://github.com/CQCL/pytket-qiskit/issues/414

I would suggest to merge this now as it is. Any objections?

cqc-melf (approved):

PR #404, opened 2024-10-23
refactor: minor improvement to `qiskit_to_tk` conversion.

CalMacCQ:
# Description

Just pulled out the circbox building in `qiskit_to_tk` into a separate function. This reduces indentation and the number of lines in the `CircuitBuilder.add_qiskit_data` method.

# Related issues
#313 

# Checklist

### DIFF
--- a/pytket/extensions/qiskit/qiskit_convert.py
+++ b/pytket/extensions/qiskit/qiskit_convert.py
@@ -524,7 +524,7 @@ class CircuitBuilder:
                 q_ctrl_box = _get_qcontrol_box(c_gate=instr, params=params)
                 self.tkc.add_qcontrolbox(q_ctrl_box, qubits)
 
-            elif isinstance(instr, (Initialize, StatePreparation)):
+            elif optype == OpType.StatePreparationBox:
                 # Append OpType found by stateprep helpers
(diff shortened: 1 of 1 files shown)


pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -524,7 +534,7 @@ def add_qiskit_data(
                 q_ctrl_box = _get_qcontrol_box(c_gate=instr, params=params)
                 self.tkc.add_qcontrolbox(q_ctrl_box, qubits)
 
-            elif isinstance(instr, (Initialize, StatePreparation)):
+            elif optype == OpType.StatePreparationBox:
```
CalMacCQ, line 537: Seems more correct to match the optype as both of the qiskit instructions can be converted to a `StatePreparationBox` instance.

CalMacCQ: In fact we could probably rewrite `CircuitBuilder.add_qiskit_data` to just check the `optype` and not look at the qiskit `Instruction`. Its just the handling of `UnitaryGate` and `PauliEvolutionGate` that are preventing this at the moment.

cqc-melf (approved):

PR #412, opened 2024-11-07
fix: Warn user about implicit qubit permutations in `tk_to_qiskit`

### LINKED ISSUE #411: Qiskit Conversion ignores implict permutations
If a `Circuit` object contains `implicit_permutations` the conversion using `tk_to_qiskit` ignores these permutations leading to an inequivalent circuit (see example below).

I would suggest to (at least) raise a warning that the conversion function is called on a circuit containing implicit permutations.

```Python
from pytket import Circuit
from pytket.extensions.qiskit import qiskit_to_tk, tk_to_qiskit
from pytket.utils import compare_statevectors

circuit_dict = {'bits': [],
     'commands': [
         {'args': [['q', [0]]], 'op': {'type': 'X'}},
         {'args': [['q', [1]]], 'op': {'type': 'Y'}},
         {'args': [['q', [2]]], 'op': {'type': 'Z'}},],
         'implicit_permutation': [[['q', [0]], ['q', [0]]],
                                  [['q', [1]], ['q', [2]]],
                                  [['q', [2]], ['q', [1]]]],
         'phase': '0.0',
         'qubits': [['q', [0]], ['q', [1]], ['q', [2]]]
    }

circ = Circuit.from_dict(circuit_dict)

# Pass the circuit through qiskit conversion
qiskit_circ = tk_to_qiskit(circ)
print(qiskit_circ)
circ_after_conversion = qiskit_to_tk(qiskit_circ)

# This statevector comparison fails
assert compare_statevectors(circ_after_conversion.get_statevector(), circ.get_statevector())
```

CalMacCQ:
# Description

Minor change to `tk_to_qiskit`. The user now gets a warning if the input pytket `Circuit` contains implicit qubit permutations.

closes #411 

# Related issues

#411 

# Checklist

### DIFF
--- a/pytket/extensions/qiskit/qiskit_convert.py
+++ b/pytket/extensions/qiskit/qiskit_convert.py
@@ -856,7 +856,7 @@ _protected_tket_gates = (
 supported_gate_rebase = AutoRebase(_protected_tket_gates)
 
 
-def _has_implicit_perumtation(circ: Circuit) -> bool:
+def _has_implicit_permutation(circ: Circuit) -> bool:
     impl_dict: dict[Qubit, Qubit] = circ.implicit_qubit_permutation()
     return tuple(impl_dict.keys()) != tuple(impl_dict.values())
 
@@ -880,7 +880,7 @@ def tk_to_qiskit(
     if replace_implicit_swaps:
         tkc.replace_implicit_wire_swaps()
 
-    if _has_implicit_perumtation(tkcirc) and not replace_implicit_swaps:
+    if _has_implicit_permutation(tkcirc) and not replace_implicit_swaps:
         warnings.warn(
             "The pytket Circuit contains implicit qubit permutations"
             + " which aren't handled by default."
--- a/pytket/extensions/qiskit/qiskit_convert.py
+++ b/pytket/extensions/qiskit/qiskit_convert.py
@@ -15,6 +15,7 @@
 
 """Methods to allow conversion between Qiskit and pytket circuit classes
 """
+import warnings
 from collections import defaultdict
 from collections.abc import Iterable
 from inspect import signature
@@ -90,7 +91,6 @@ from qiskit.circuit.library import (
     StatePreparation,
     UnitaryGate,
 )
-import warnings
 
 if TYPE_CHECKING:
     from qiskit_ibm_runtime.ibm_backend import IBMBackend  # type: ignore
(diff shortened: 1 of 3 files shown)


tests/qiskit_convert_test.py
```diff
@@ -1172,3 +1172,10 @@ def test_nonregister_bits() -> None:
     c.rename_units({Bit(0): Bit(1)})
     with pytest.raises(NotImplementedError):
         tk_to_qiskit(c)
+
+
+def test_implicit_swap_warning() -> None:
+    c = Circuit(2).H(0).SWAP(0, 1)
+    c.replace_SWAPs()
+    with pytest.warns(UserWarning, match="The pytket Circuit contains implicit qubit"):
```
CalMacCQ, line 1180: Kindof suprised that python can use this `UserWarning` class without it being imported.

pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -855,6 +856,11 @@ def append_tk_command_to_qiskit(
 supported_gate_rebase = AutoRebase(_protected_tket_gates)
 
 
+def _has_implicit_permutation(circ: Circuit) -> bool:
```
cqc-alec, line 859: This would be better defined in pytket rather than this extension: there's nothing specific to qiskit about it. But for now there is no need to factor it out as a method, since it's only used in one place.

pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -855,6 +856,11 @@ def append_tk_command_to_qiskit(
...
+    impl_dict: dict[Qubit, Qubit] = circ.implicit_qubit_permutation()
+    return tuple(impl_dict.keys()) != tuple(impl_dict.values())
```
cqc-alec, line 861:
```suggestion
    return any(q0 != q1 for q0, q1 in circ.implicit_qubit_permuation().items())
```
More efficient because of shortcutting.

cqc-alec (changes requested):

pytket/extensions/qiskit/qiskit_convert.py
CalMacCQ: Thanks

pytket/extensions/qiskit/qiskit_convert.py
CalMacCQ: Agreed. Maybe it should just be a property of `Circuit`?

cqc-alec: Yes. But for the purpose of this PR the check can just be inlined.

cqc-alec (approved):

PR #423, opened 2024-11-25
Overloaded `get_compiled_circuit/s`

sjdilkes:
The timeout wasn't being passed appropriately in the previous version. This PR overloads the `Backend.get_compiled_circuit` and `Backend.get_compiled_circuits` methods to add a `timeout` parameter, as with `pytket-quantinuum`, making sure users can modify it for using optimisation level 3.

# Checklist


pytket/extensions/qiskit/backends/ibm.py
```diff
@@ -410,7 +410,34 @@ def default_compilation_pass_offline(
             passlist.append(FullPeepholeOptimise())
         elif optimisation_level == 3:
             passlist.append(RemoveBarriers())
-            passlist.append(AutoRebase({OpType.CX, OpType.H, OpType.Rz}))
+            passlist.append(
```
cqc-melf, line 413: Is this a relevant change, that should be in the changelog, too?

pytket/extensions/qiskit/backends/aer.py
```diff
@@ -256,6 +309,75 @@ def default_compilation_pass(
...
+        :param timeout: Only valid for optimisation level 3, gives a maximimum time
+            for running a single thread of the pass `GreedyPauliSimp`. Increase for
+            optimising larger circuits.
+        :type timeout: int, optional
+
+        :return: An optimised quantum circuit
+        :rtype: Circuit
+        """
+        return_circuit = circuit.copy()
+        if optimisation_level == 3 and circuit.n_gates_of_type(OpType.Barrier) > 0:
```
cqc-melf, line 331: Why is this check not in the other function `get_compiled_circuits` ?

pytket/extensions/qiskit/backends/ibm.py
```diff
@@ -441,6 +468,75 @@ def default_compilation_pass_offline(
...
+        :param timeout: Only valid for optimisation level 3, gives a maximimum time
+            for running a single thread of the pass `GreedyPauliSimp`. Increase for
+            optimising larger circuits.
+        :type timeout: int, optional
+
+        :return: An optimised quantum circuit
+        :rtype: Circuit
+        """
+        return_circuit = circuit.copy()
+        if optimisation_level == 3 and circuit.n_gates_of_type(OpType.Barrier) > 0:
```
cqc-melf, line 490: See above

pytket/extensions/qiskit/backends/aer.py
sjdilkes: `get_compiled_circuits` calls this method so the check is done implicitly

pytket/extensions/qiskit/backends/ibm.py
sjdilkes: same answer as for `AerBackend`

pytket/extensions/qiskit/backends/ibm.py
sjdilkes: I don't think we don't need to include this - all I've done is update the set of gates we `AutoRebase` to so that it exactly matches what `GreedyPauliSimp` accepts. We don't mention this rebase for the `GreedyPauliSimp` line of the changelog so I don't think it's worth explicitly pointing out.

cqc-melf (approved):

Issue #366, opened 2024-07-16
Replace bit condition “c_if” in bit-string condition “if_test” of "qiskit_convert.py"


yoshi-qc:
In “tk_to_qiskit”, a conditional gate in TKET circuit transpires a conditional gate of qiskit circuit using bit condition “c_if” on L660-L697 of “qiskit_convert.py”.
https://github.com/CQCL/pytket-qiskit/blob/main/pytket/extensions/qiskit/qiskit_convert.py
https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.InstructionSet#c_if
I found that bit-string condition “if_test” is currently supported and recommended in qiskit. Should we edit the lines L660-L697 of “qiskit_convert.py”?
https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.QuantumCircuit#if_test

https://github.com/CQCL/pytket-qiskit/blob/94174692d5135a3fd9b55b9a3b99ec53ad342942/pytket/extensions/qiskit/qiskit_convert.py#L660

yoshi-qc:
The conditional phase is supported in qiskit circuit.
https://github.com/CQCL/pytket-qiskit/blob/d5b00fc96d2355c433b8885cce04d92310111266/pytket/extensions/qiskit/qiskit_convert.py#L662
For instance,
```
with qc.if_test((bits[4], 1)):
    qc.append(GlobalPhaseGate(1),[],[])
```
https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.GlobalPhaseGate

cqc-alec:
@yoshi-qc what is the priority on this? Does it affect functionality?

yoshi-qc:
I discuss this issue with @sjdilkes . He manages this issue now.

CalMacCQ:
Related to #415

cqc-alec:
Closing as duplicate. I've made some of the changes necessary on a branch (see https://github.com/CQCL/pytket-qiskit/commit/1929e60be3e076029c099c55d874acb7a36e9567 ) but it can't be merged because it causes test failures due to qiskit not having caught up yet.

PR #437, opened 2024-12-17
feat: handle `IfElseOp` in `qiskit_to_tk`

### LINKED ISSUE #415: Support  `QuantumCircuit.if_test` in `qiskit_to_tk`
```python
from qiskit import QuantumCircuit, QuantumRegister, ClassicalRegister
from pytket.extensions.qiskit import qiskit_to_tk
 
qubits = QuantumRegister(1)
clbits = ClassicalRegister(1)
circuit = QuantumCircuit(qubits, clbits)
(q0,) = qubits
(c0,) = clbits
 
circuit.h(q0)
circuit.measure(q0, c0)
with circuit.if_test((c0, 1)):
    circuit.x(q0)
circuit.measure(q0, c0)
 
tket_circ = qiskit_to_tk(circuit)
```
Errors with

```
  348     return optype
    349 except KeyError:
--> 350     raise NotImplementedError(
    351         f"Conversion of qiskit's {instruction.name} instruction is "
    352         + "currently unsupported by qiskit_to_tk. Consider "
    353         + "using QuantumCircuit.decompose() before attempting "
    354         + "conversion."
    355     )

NotImplementedError: Conversion of qiskit's if_else instruction is currently unsupported by qiskit_to_tk. Consider using QuantumCircuit.decompose() before attempting conversion.

```

on `pytket-qiskit ==0.58.0`

CalMacCQ:
# Description

Converting a qiskit [IfElseOp](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.IfElseOp) as two conditional pytket `CircBox`es. This sort of control flow is probably better expressed in the HUGR data structure than as a pytket `Circuit/CircBox`. 

Theres a C.I. failure for one of the tests I added that I can't reproduce locally -> https://github.com/CQCL/pytket-qiskit/actions/runs/12811218709/job/35720030508#step:7:786

Needs more diverse test cases before merging and we also need to decide if we want to handle more complex conditions.

NOTE: For now I've restricted to conditions on a single bit but we should generalise this. This PR is already quite busy.

Looking at the qiskit docs they also have switch-case as well as for and while loops which we may want to support in the future.

https://docs.quantum.ibm.com/guides/classical-feedforward-and-control-flow

# Example

```python
# define qiskit program
qubits = QuantumRegister(2)
clbits = ClassicalRegister(2)
circuit = QuantumCircuit(qubits, clbits)
(q0, q1) = qubits
(c0, c1) = clbits

circuit.h(q0)
circuit.measure(q0, c0)

with circuit.if_test((c0, 1)) as else_:
    circuit.h(q1)
with else_:
    circuit.x(q1)
circuit.measure(q1, c1)
```

Now the corresponding pytket circuit looks like

```python
tkc = qiskit_to_tk(circuit)
```
circuit

The `If` and `Else` boxes contain a single $H$ and $X$ gate respectively.

I thought about putting string representing the conditon inside the "If" label however this could be awkward. Hopefully its clear from the diagram whats going on.

This conversion ended up being slighly tricker than I expected because the qiskit circuits which define the `true_body` and `false_body` do not have registers. Managed to get around this by using the register from the circuit containing the `IfElseOp` and cleaning up empty wires with `Circuit.remove_blank_wires` to ensure the `CircBox` is of the correct size.

# Related issues

closes #415

# Checklist


pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -478,6 +479,35 @@ def _build_circbox(instr: Instruction, circuit: QuantumCircuit) -> CircBox:
...
+    default_qreg_if  = QuantumRegister(if_qc.num_qubits, "q")
+    default_creg_if  = ClassicalRegister(if_qc.num_clbits, "c")
+    default_qreg_else  = QuantumRegister(else_qc.num_qubits, "q")
+    default_creg_else  = ClassicalRegister(else_qc.num_clbits, "c")
+
+    new_if_qc = QuantumCircuit(default_qreg_if, default_creg_if)
+    new_else_qc = QuantumCircuit(default_qreg_else, default_creg_else)
+
+    if_builder = CircuitBuilder(new_if_qc.qregs, new_if_qc.cregs)
+    if_builder.add_qiskit_data(if_qc)
```
CalMacCQ, line 497: I think this call to the `add_qiskit_data` method is wrong.

pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -479,6 +480,51 @@ def _build_circbox(instr: Instruction, circuit: QuantumCircuit) -> CircBox:
...
+def build_if_else_circuit(
+    if_else_op: IfElseOp,
+    qregs: list[QuantumRegister],
+    cregs: list[ClassicalRegister],
+    qubits: list[Qubit],
+    bits: list[Bit],
+) -> Circuit:
+    if_box, else_box = _pytket_boxes_from_IfElseOp(if_else_op, qregs, cregs)
+    circ_builder = CircuitBuilder(qregs, cregs)
+    circ = circ_builder.circuit()
```
yao-cqc, line 511:
```
print(circ.qubits, circ.bits)
print(if_box.get_circuit().qubits, if_box.get_circuit().bits)
print(qubits)
print(bits)
```
prints
```
[q105[0], q105[1]] [c3[0], c3[1]]
[q105[0], q105[1]] [c3[0], c3[1]]
[q105[1]]
[c3[0]]
```

So the box has 2 qubits and 2 bits, but in `add_circbox`, `args=qubits + bits` only provides 1 qubit and 1 bit.

CalMacCQ:
The test case I added is now working.

I'll need to add more diverse test cases before I'm confident in merging this as well as negate the "else" condition properly and see what the story is for more complex conditions.

pytket/extensions/qiskit/qiskit_convert.py
CalMacCQ: Thanks. 

I figured out what I was doing wrong. The thing that was confusing me was that the two circuits which define the `IfElseOp` do not have any registers. Their qubits are tied to the register of the circuit containing the `IfElseOp` this meant that the subcircuits in my `CircBox`es had too many qubits/bits as there were some additional blank wires.

Cleaning up the blank wires as in https://github.com/CQCL/pytket-qiskit/pull/437/commits/d65a347292a0238499dde5a10abb6442543ec745 seems to fix this.

pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -479,6 +480,69 @@ def _build_circbox(instr: Instruction, circuit: QuantumCircuit) -> CircBox:
     return CircBox(subc)
 
 
+# Used for handling of IfElseOp
+# docs -> https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.IfElseOp
+# Examples -> https://docs.quantum.ibm.com/guides/classical-feedforward-and-control-flow
+def _pytket_boxes_from_ifelseop(
+    if_else_op: IfElseOp, qregs: list[QuantumRegister], cregs: list[ClassicalRegister]
+) -> tuple[CircBox, Optional[CircBox]]:
```
CalMacCQ, line 488: I'm not so pleased with this `Optional` return type in the function signature. This is to deal with the case when the `IfElseOp` only has a `true_body` rather than a `true_body` and a `false_body`.

There's probably a cleaner way to deal with this.

tests/qiskit_convert_test.py
```diff
@@ -1193,6 +1193,63 @@ def test_nonregister_bits() -> None:
...
+    circuit.measure(q0, c0)
+
+    with circuit.if_test((c0, 1)) as else_:
+        circuit.h(q1)
+    with else_:
+        circuit.x(q1)
+    circuit.measure(q1, c1)
+
+    tkc = qiskit_to_tk(circuit)
+    assert tkc.n_gates_of_type(OpType.Conditional) == 2
```
CalMacCQ, line 1213: Should test this more thoroughly. Annoyingly ran into this bug -> https://github.com/CQCL/tket/issues/1726 when trying to validate the contents of the two `CircBox`es.

pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -479,6 +480,69 @@ def _build_circbox(instr: Instruction, circuit: QuantumCircuit) -> CircBox:
...
+        condition_value=if_else_op.condition[1],
+    )
+    # If we have an else_box defined, add it to the circuit
+    if else_box is not None:
+        circ.add_circbox(
+            circbox=else_box,
+            args=qubits,
+            condition_bits=bits,
+            # TODO: handle conditions over multiple bits/registers?
+            condition_value=not bool(if_else_op.condition[1]),
```
CalMacCQ, line 541: Still need to decide if its worth handling more complex conditionals here or just restrict to conditions over a single bit for now.

tests/qiskit_convert_test.py
```diff
@@ -1193,6 +1193,80 @@ def test_nonregister_bits() -> None:
...
+
+    exp_if_circ = Circuit()
+    r = exp_if_circ.add_q_register("r", 1)
+    exp_if_circ.H(r[0])
+
+    exp_else_circ = Circuit()
+    s = exp_else_circ.add_q_register("s", 1)
+    exp_else_circ.H(s[0])
+
+    assert if_circ == exp_if_circ
```
CalMacCQ, line 1228: something weird is going on with register names here so this. equality check fails.

`if_circ`
if1

`exp_if_circ`
if2

CalMacCQ: Ah looking more closely at the original circuit this is actually expected as the Hadamard is applied to qubit one.

Creating a fresh pytket circuit the register starts from zero so thats why the comparison fails (expected)

I've added a simpler gate counting test for now.

pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -480,6 +481,77 @@ def _build_circbox(instr: Instruction, circuit: QuantumCircuit) -> CircBox:
...
+    # Get two CircBox objects which implement the true_body and false_body.
+    if_box, else_box = _pytket_boxes_from_ifelseop(if_else_op, qregs, cregs)
+    # else_box can be None if no false_body is specified.
+
+    circ_builder = CircuitBuilder(qregs, cregs)
+    circ = circ_builder.circuit()
+
+    # Coniditions must be on a single bit (for now)
+    assert len(if_else_op.condition) == 2
+    if not isinstance(if_else_op.condition[0], Clbit):
```
CalMacCQ, line 532: As I said in the description I've restricted to the case of single bit conditions for now. However it should be easy enough to generalise for conditons on values of entire registers in a follow up PR.

tests/qiskit_convert_test.py
```diff
@@ -1193,6 +1193,93 @@ def test_nonregister_bits() -> None:
...
+    x_circ.name = "If"
+    xq1 = x_circ.add_q_register("q1", 1)
+    x_circ.X(xq1[0])
+    expected_circ.add_circbox(
+        CircBox(x_circ), [q1_tk[0]], condition_bits=[c1_tk[0]], condition_value=1
+    )
+
+    expected_circ.Measure(q1_tk[0], c1_tk[0])
+
+    assert tket_circ_if_else == expected_circ
```
cqc-alec, line 1259: A similar test including an `else` condition would be good to add.

cqc-alec (changes requested):
The PR description says that this closes #366 , but this does not change `tk_to_qiskit` a all does it? Did you intend to address that direction as well?

pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -480,6 +481,77 @@ def _build_circbox(instr: Instruction, circuit: QuantumCircuit) -> CircBox:
...
+) -> Circuit:
+    # Get two CircBox objects which implement the true_body and false_body.
+    if_box, else_box = _pytket_boxes_from_ifelseop(if_else_op, qregs, cregs)
+    # else_box can be None if no false_body is specified.
+
+    circ_builder = CircuitBuilder(qregs, cregs)
+    circ = circ_builder.circuit()
+
+    # Coniditions must be on a single bit (for now)
+    assert len(if_else_op.condition) == 2
```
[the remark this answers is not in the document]
CalMacCQ: Yeah I considered this. I should add an exception to do things properly. 

My assumption was that the object returned from `if_else_op.condition` would always be of length 2 however this may not be true.

 I think for conditions on multiple bits this could be longer.

pytket/extensions/qiskit/qiskit_convert.py
CalMacCQ: I prefer the former suggestion as the restriction to a single bit should be removable soon

pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -480,6 +481,77 @@ def _build_circbox(instr: Instruction, circuit: QuantumCircuit) -> CircBox:
...
+        condition_value=if_else_op.condition[1],
+    )
+    # If we have an else_box defined, add it to the circuit
+    if else_box is not None:
+        circ.add_circbox(
+            circbox=else_box,
+            args=qubits,
+            condition_bits=bits,
+            # TODO: handle conditions over multiple bits/registers?
+            condition_value=not bool(if_else_op.condition[1]),
```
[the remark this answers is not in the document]
CalMacCQ: Added a check in [cebf4dd](https://github.com/CQCL/pytket-qiskit/pull/437/commits/cebf4ddd35537878f183193618c96adc7a025b0b)

tests/qiskit_convert_test.py
CalMacCQ: Turned the two branch test into an explicit equality check in [a40732d](https://github.com/CQCL/pytket-qiskit/pull/437/commits/a40732db1d07c8eb190090cc3938db867b1c35b5)

pytket/extensions/qiskit/qiskit_convert.py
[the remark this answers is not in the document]
CalMacCQ: Redundant now as I've redone the test

cqc-alec (approved):

PR #459, opened 2025-02-03
docs: fix silently failing c.i. build of the sphinx docs

### LINKED ISSUE #458: pytket-qiskit docs C.I. build failing silently
The reason for the docs build failure is the fact that the line in [build_and_test.yml](https://github.com/CQCL/pytket-qiskit/blob/main/.github/workflows/build_and_test.yml#L83C9-L83C96) is failing to install pytket-qiskit.

```
for w in `find wheelhouse/ -type f -name "*.whl"` ; do poetry run pip install $w ; done
```

This wasn't showing up as no exit flags were added to the `build-docs.sh` script. Hence the silent build failure

Note that this doesn't impact the website build but would be nice to have C.I. checks running properly.

See #443 

@cqc-melf Also likely to be an issue with other extensions because the docs build c.i. checks are very similar. Worth checking.

CalMacCQ:
closes #458 

The issue was twofold. Firstly the `wheelhouse` directory could not be found after the "cd" into the docs directory. secondly we were also not using "recursive" mode for the docs submodules in c.i.

Will add link checking in a follow up PR I think. Its proving to be a bit more annoying than I thought to get it working with the other stuff in `build-docs.sh`

I think we should maybe change a few things about how this works going forward.


CalMacCQ:
Here's the C.I. failure

Screenshot 2025-02-03 at 13 05 16

This didn't manifest before for two reasons.
1. Sphinx warnings were allowed by the build. We can treat warninngs as errors with the `-W` flag in `sphinx-build`
2. Even if warnings were treated as errors, the `build-docs.sh` script was missing exit flags so the errors would not fail the C.I. checks.

.github/workflows/build_and_test.yml
```diff
@@ -79,8 +79,10 @@ jobs:
     - name: Install docs dependencies
       if:  (matrix.os == 'ubuntu-latest') && (github.event_name == 'pull_request' || github.event_name == 'schedule' )
       run: |
-        cd docs && bash ./install.sh
-        for w in `find wheelhouse/ -type f -name "*.whl"` ; do poetry run pip install $w ; done
+        bash ./docs/install.sh
+        cd docs
+        for w in `find ../wheelhouse/ -type f -name "*.whl"` ; do poetry run pip install $w ; done
```
CalMacCQ, line 84: Seems the reason this wasn't working it that we couldn't find the `wheelhouse` directory after "cd"ing into docs

docs/build-docs.sh
```diff
@@ -1,20 +1,32 @@
 #!/bin/bash
+set -e
 rm -rf build/
 
+# This build script is only used for local docs build.
```
cqc-melf, line 5: And on the CI for doc checks?

cqc-melf (commented):
What are the implications for the other extension repos? Should we update them as well?

docs/build-docs.sh
CalMacCQ: Good point, I'll clarify that.

PR #482, opened 2025-04-30
Update version and changelog for release.

cqc-alec:


docs/changelog.md
```diff
@@ -4,7 +4,7 @@
 
 # Changelog
 
-## Unreleased
+## 0.67.0 (April 2025)
 
 - Add support for Python 3.13.
 - Support qiskit >=2.0, drop support for qiskit <2.0.
```
CalMacCQ, line 10: Maybe we should say explicitly that pytket condtional gates are now handled as `IfElseOp` instances in qiskit?

cqc-alec: I think that's a detail that doesn't need to go in the changelog.

PR #487, opened 2025-05-19
Update version and changelog for 0.68.0 release.

cqc-alec:


CalMacCQ:
Should we add whatever the VF2 pass is to the table here?

https://docs.quantinuum.com/tket/extensions/pytket-qiskit/#default-compilation

docs/changelog.md
```diff
@@ -4,7 +4,7 @@
 
 # Changelog
 
-## Unreleased
+## 0.68.0 (May 2025)
 
 - Add `VF2PostLayout` to default compilation.
```
cqc-melf, line 9: Should we update the pytket version before doing the release?

cqc-melf (dismissed):

docs/changelog.md
cqc-alec: May as well do that yes...

PR #509, opened 2025-07-17
Add docs coverage and warning checks to CI and resolve failures for both

willsimmons1465:
# Description

Adding CI checks for documentation coverage (making sure that every publicly visible method and class - those whose names don't begin with an underscore - is included in the docs) and enforcing sphinx nitpicky to report any formatting failures and broken cross-references.

This PR also resolves existing gaps in coverage and broken formatting. Some of the newly documented content required for docs coverage may have been intended to be internal components required only as part of the conversion or infrastructure code and is not intended for end users; I will need some help spotting any such instances so we can underscore them out to clarify this intention.

# Related issues

This works towards CQCL/tket#1857, following on from CQCL/tket#1910 for the core package and related PRs for other extensions.

# Checklist


docs/index.md
```diff
@@ -32,16 +32,16 @@ and methods into the `pytket.extensions` namespace.
 .. autosummary::
     :nosignatures:
 
-    IBMQBackend
-    IBMQEmulatorBackend
-    AerBackend
-    AerStateBackend
-    AerUnitaryBackend
-    AerDensityMatrixBackend
+    backends.ibm.IBMQBackend
```
CalMacCQ, line 35: Could we prefix these with `~` to make the table look slightly nicer?

Screenshot 2025-07-24 at 15 01 02

docs/api.md
```diff
@@ -56,7 +104,7 @@
 
 ```{eval-rst}
 .. automodule:: pytket.extensions.qiskit.backends.crosstalk_model
-    :members: CrosstalkParams
+    :members: CrosstalkParams, NoiseGate, NoisyCircuitBuilder, FractionalUnitary, get_gate_times_from_backendinfo
```
CalMacCQ, line 107: @yao-cqc Do you know which of these classes are intended to be public?

docs/api.md
```diff
@@ -1,50 +1,98 @@
...
+.. autoclass:: pytket.extensions.qiskit.backends.aer.NoiseModelCharacterisation
+    :members:
+```
+
+```{eval-rst}
+.. automodule:: pytket.extensions.qiskit.qiskit_convert
+
+    .. autofunction:: qiskit_to_tk
+    .. autofunction:: tk_to_qiskit
+    .. autofunction:: process_characterisation
```
CalMacCQ, line 90: I think `param_to_qiskit` and `param_to_tk` can probably be underscored. They are just helpers for the circuit conversion code.

.github/workflows/build_and_test.yml
```diff
@@ -32,7 +32,7 @@ jobs:
     - uses: actions/checkout@v4
       with:
         fetch-depth: '0'
-        submodules: true
+        submodules: recursive
```
CalMacCQ, line 35: Technically this wouldn't be needed anymore as we no longer have a submodule that contains a submodule.

`quantinuum-sphinx` is now installable from the github repo rather than a submodule of `pytket-docs-theming`.

CalMacCQ (commented):
Thanks for doing this :)  A few comments

docs/api.md
willsimmons1465: Will do. Btw, `get_gate_times_from_backendinfo` is not used anywhere, can I remove it entirely?

.github/workflows/build_and_test.yml
willsimmons1465: I don't see any particular downside to keeping it recursive until someone adds something huge as a submodule, which would likely not be a good thing to do anyway; but I don't feel very strongly either way

CalMacCQ (approved):

PR #515, opened 2025-08-24
implemented support for Var, Unary, and Binary operations during `qiskit_to_tk` conversion

### LINKED ISSUE #514: Incorrect bits list referenced in `_append_if_else_circuit`
Consider the conversion of the following Qiskit QuantumCircuit to a pytket Circuit:

```py
from qiskit import QuantumRegister, ClassicalRegister, QuantumCircuit

qreg = QuantumRegister(1)
qreg_setter = QuantumRegister(2)
creg_A = ClassicalRegister(1)
creg_B = ClassicalRegister(1)

qc = QuantumCircuit(qreg, qreg_setter, creg_A, creg_B)
qc.x(qreg_setter[1])

with qc.if_test((creg_A[0], 0)) as _else:
    qc.measure(qreg_setter[1], creg_B[0])
with _else:
    qc.measure(qreg_setter[0], creg_B[0])
```

The `IfElseOp` in this circuit only has a single classical bit in its condition, but `_append_if_else_circuit` sees the `bits` argument, which includes the bit `creg_B[0]` which is _inside the operation_, _not inside the condition_, and thus raises the following error:
```py
NotImplementedError: Conditions on multiple bits not supported
```

The function should instead be looking at the bits of `if_else_op.condition` to allow `IfElseOp`'s which act on classical bits. This would also affect the creation of the `CircBox` itself, which takes `bits` to be the condition_bits.

RasmitDevkota:
# Description

Implements support for Var, Unary, and Binary operations in `_append_if_else_circuit` during `qiskit_to_tk` conversion.

# Related issues

Closes https://github.com/CQCL/pytket-qiskit/issues/514

### DIFF
--- a/pytket/extensions/qiskit/qiskit_convert.py
+++ b/pytket/extensions/qiskit/qiskit_convert.py
@@ -689,9 +689,9 @@ def _append_if_else_circuit(
             # Find the pytket Bit with the same register name and index as the Qiskit Clbit
             for bit in bits:
                 if (
-                    bit.reg_name == _condition.var._register.name
-                    and bit.index[0] == _condition.var._index
-                ):  # noqa: SLF001
+                    bit.reg_name == _condition.var._register.name  # noqa: SLF001
+                    and bit.index[0] == _condition.var._index  # noqa: SLF001
+                ):
                     return bit
 
             raise ValueError(
@@ -704,9 +704,9 @@ def _append_if_else_circuit(
             # Find the pytket Bit with the same register name and index as the Qiskit Clbit
             for bit in bits:
                 if (
-                    bit.reg_name == _condition.operand.var._register.name
-                    and bit.index[0] == _condition.operand.var._index
-                ):  # noqa: SLF001
+                    bit.reg_name == _condition.operand.var._register.name  # noqa: SLF001
+                    and bit.index[0] == _condition.operand.var._index  # noqa: SLF001
+                ):
                     return bit ^ val
 
             raise ValueError(
@@ -755,9 +755,9 @@ def _append_if_else_circuit(
         condition_bits = [
             bit
             for bit in bits
-            if bit.reg_name == if_else_op.condition.var._register.name
-            and bit.index[0] == if_else_op.condition.var._index
-        ]  # noqa: SLF001
+            if bit.reg_name == if_else_op.condition.var._register.name  # noqa: SLF001
+            and bit.index[0] == if_else_op.condition.var._index  # noqa: SLF001
+        ]
 
         if len(condition_bits) == 0:
             raise ValueError(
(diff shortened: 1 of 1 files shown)


CalMacCQ:
Thanks a lot for the PR :) 

We can take a detailed look shortly. 

A couple things 
1. There are some linting issues [here](https://github.com/CQCL/pytket-qiskit/actions/runs/17194545170/job/48900952337?pr=515) Most of these should be easy to address. If its not possible to avoid the private attribute access then you can add a NOQA comments as is done [here](https://github.com/CQCL/pytket-qiskit/blob/b87c8ca87d6b7d54cbbc811a1eade7b4a3f906f2/pytket/extensions/qiskit/tket_pass.py#L60).
2. Could you please add a test which would have failed before this change but now passes? You can add this to `test_qiskit_convert.py`. Maybe the snippet you showed in the issue would be a good building block for a test.

CalMacCQ:
Regarding the lints I think that `ruff check . --fix` should sort out the rest? I could be wrong.

RasmitDevkota:
Looks like I had to ask ruff to do extra checks to get through the rest of the lint errors, but I think that should cover it.

CalMacCQ:
I think after reformatting the NOQA comments are now on the wrong line which is kinda annyoying.

Would also be nice to have a test in `qiskit_convert_test.py` to demonstrate the fix and guard against regressions with future changes to the conversion code.

CalMacCQ:
> Working on that now, would the standard structure be to test each of the different types of conditions separately, as in having a separate test for Var, Unary, and Binary, or something more compact? I have examples which reach all of the new branches I've created but I'm just not sure about organization.

Regarding the structure of the tests I don't have a strong opinion really. It would be desirable if the behaviour for `Var`, `Unary` and `Binary` was tested. Apart from that, feel free to write the test in a way that you're comfortable with and one of us can review and provide feedback if needed.

Also spotted that the linting is fixed now, thanks!

CalMacCQ:
Happy to help more with this if needed. Thanks again for the contribution.

CalMacCQ:
> Sorry for the late response - some help would definitely be appreciated! I've just been struggling to construct the pytket circuit correctly so that the circuit objects would be equal in the end.

I'm going on leave for a few days but I'll get back to you on Monday.

CalMacCQ:
Looking at this now :)

cqc-melf:
@CalMacCQ what is the state of this? Do you need help?

CalMacCQ:
I split off the fix that @RasmitDevkota made into a separate PR. https://github.com/Quantinuum/pytket-qiskit/pull/524 This change was merged and released some time ago. 

I personally didn't have plans to work on this anymore. If someone can make a case that these are useful conversions to support and maintain then we can maybe come back to it.

cqc-melf:
Please reopen and ping me if you think this is still relevant.

PR #524, opened 2025-09-30
fix: address the condition bits in handling of `IfElseOp`

### LINKED ISSUE #514: Incorrect bits list referenced in `_append_if_else_circuit`

CalMacCQ:
# Description

In this PR I am splitting off the fix for #514 from the rest of the changes made in #515 for ease of testing. 

I added a test case based on code that was provided in #514 and veified that it failed. I then added the fix originally implemented by @RasmitDevkota.

Note that we could still merge  the `Var` , `Unary` and `Binary`  conditions implemented in #515 

# Related issues

closes #514 

# Checklist


pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -676,32 +677,82 @@ def _append_if_else_circuit(
...
+        )
+        # If we have an else_circ defined, add it to the circuit
+        if else_circ is not None:
+            outer_builder.tkc.add_circbox(
+                circbox=CircBox(else_circ),
+                args=else_circ.qubits + else_circ.bits,  # type: ignore
+                condition_bits=condition_bits,
+                condition_value=0,
+            )
+    elif hasattr(if_else_op.condition, "__getitem__") and isinstance(
```
CalMacCQ, line 719: Maybe checking whether `if_else_op.condition` is an instance of `Tuple` makes more sense?

pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -676,32 +677,82 @@ def _append_if_else_circuit(
...
+            raise ValueError(
+                "Failed to find any pytket Bit matching Qiskit Clbit in condition for IfElseOp."
+            )
+
+        # In this case, if_else_op.condition is a single tuple of shape (Clbit, value)
+        outer_builder.tkc.add_circbox(
+            circbox=CircBox(if_circ),
+            args=if_circ.qubits + if_circ.bits,  # type: ignore
+            condition_bits=condition_bits,
+            condition_value=1,
```
cqc-alec, line 709: If there are multiple condition_bits, this just tests that one of them is set and the others are unset. Don't we want to test that they are all set?

pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -676,32 +677,82 @@ def _append_if_else_circuit(
...
+        )
+        # If we have an else_circ defined, add it to the circuit
+        if else_circ is not None:
+            outer_builder.tkc.add_circbox(
+                circbox=CircBox(else_circ),
+                args=else_circ.qubits + else_circ.bits,  # type: ignore
+                condition_bits=condition_bits,
+                condition_value=0,
```
cqc-alec, line 717: Similarly: this will be true iff all the condition bits are unset. I think the condition should be that any of them is unset?

pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -716,7 +743,7 @@ def _append_if_else_circuit(
     else:
         raise TypeError(
```
cqc-alec, line 744: According to the [qiskit docs](https://quantum.cloud.ibm.com/docs/en/api/qiskit/qiskit.circuit.IfElseOp) this can also be an `expr.Expr`, so perhaps we should assert that and raise a `NotImplementedError` for this case.

pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -676,32 +677,58 @@ def _append_if_else_circuit(
...
-    if isinstance(if_else_op.condition[0], Clbit):
-        if len(bits) != 1:
-            raise NotImplementedError("Conditions on multiple bits not supported")
+
+    if isinstance(if_else_op.condition, (Unary | Binary | Var)):
+        raise NotImplementedError(
+            "Handling of Unary, Var and Binary conditions is not supported yet."
+        )
+
+    if hasattr(if_else_op.condition, "__getitem__") and isinstance(
```
cqc-alec, line 690: Probably better to test for it being a `tuple`, which is how it's documented.

pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -676,32 +677,58 @@ def _append_if_else_circuit(
...
+        if_else_op.condition[0], Clbit
```
cqc-alec, line 691: We could also check first if `if_else_op.condition` is empty, but that's probably unlikely and fine to error out on access.

pytket/extensions/qiskit/qiskit_convert.py
```diff
@@ -676,32 +677,58 @@ def _append_if_else_circuit(
...
+        ]
+
+        if len(condition_bits) == 0:
+            raise ValueError(
+                "Failed to find any pytket Bit matching Qiskit Clbit in condition for IfElseOp."
+            )
+        if len(condition_bits) > 1:
+            raise ValueError(
+                f"Found condition on {len(condition_bits)},"
+                "only conditions on single individual bits are supported at present."
```
cqc-alec, line 707: I guess we'll only hit this error if `bits` contains duplicates (or multi-indexed bits), so perhaps the error message should say this explicitly.

pytket/extensions/qiskit/qiskit_convert.py
CalMacCQ: Good point. I've added it here to be sure [6e5eeac](https://github.com/CQCL/pytket-qiskit/pull/524/commits/6e5eeac99fe5999e39a634b924a7939857055c52). 

Its not super clear from the source but I was under the impression that `Var`, `Binary` and `Unary` inherit from `Expr`. Can't hurt to check for it explicitly though.

cqc-melf (dismissed):

cqc-melf:
The docs are still having issues, besides that it looks good :)

cqc-melf (approved):

